PR #24, opened 2026-07-16
Register remaining regression tests with CTest

Dafenxz0:
## What changed

- Register the four regression executables that the current CTest configuration does not yet expose.
- Add a `kyty_tests` target that builds all twelve regression executables before running CTest.
- Document the aggregate build and test commands for Windows and Linux.

## Why

`main` includes CTest and several focused test entries, but four existing regression executables still have to be discovered and run manually. This completes the registration without replacing the newer per-area tests that landed after this PR was opened.

## Rebase resolution

The branch is rebased onto `e63f5b7`. I kept the current `main` registrations for `scalar_provenance` and `virtual_memory_allocation`, retained the newer specialized CTest entries and macOS documentation, and added only the four missing executables to avoid duplicate test names.

## Validation

- Confirmed all 24 registered CTest names are unique.
- Confirmed every dependency of `kyty_tests` names an executable target defined in `src/CMakeLists.txt`.
- Ran `git diff --check` after the rebase.
- A full current-tree configure and build was not available in the isolated worktree because Ninja and `Qt6_DIR` were not configured.

## AI assistance disclosure

AI assistance was used to inspect the updated CMake configuration and resolve the rebase. The resulting two-file diff and validation results were reviewed before submission.

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,17 @@ codesign --force --sign - _Build/macos/install/libMoltenVK.dylib
 
 Release archives already include a signed `libMoltenVK.dylib`.
 
+### Regression tests
+
+Build every regression executable and run the registered tests with:
+
+```powershell
+cmake --build _Build/windows --target kyty_tests
+ctest --test-dir _Build/windows --output-on-failure
+```
+
+Use `_Build/linux` instead of `_Build/windows` for a Linux build.
+
 ### Visual Studio Code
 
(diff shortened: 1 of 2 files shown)


Nmzik (approved):
Reviewed the complete change. The CTest registration, aggregate test target, documentation, and deterministic UNORM test values are consistent; no actionable issues found.

PR #28, opened 2026-07-17
Fix AMD RDNA startup crash and rendering corruption

kirathenotebook:
# What changed
- Drop the unused VK_FORMAT_D24_UNORM_S8_UINT device check that rejected some AMD GPUs.
- Only treat VALIDATION-type Vulkan messages as fatal, not GENERAL-type ones.
- Restrict compute queues to the graphics queue's family.
- Force per-invocation lane mask emulation for 64-wide waves, since RDNA's lane layout differs from GCN despite matching wave size.

# Why
On AMD Radeon GPUs the emulator failed to find a suitable device, crashed during queue setup, or rendered corrupted output, while NVIDIA was unaffected.

Each change removes an assumption in device selection, debug message handling, queue selection, or shader lane-mask emulation that happened to hold for NVIDIA but not for AMD.

# Validation
- Built on Windows with clang-cl. Note: the latest Vulkan SDK (1.4.350) fails to build against the pinned Vulkan-Headers submodule; used 1.4.328.1 instead, which matches it.
- Tested on a Radeon RX 9070 XT + Ryzen 7 9800X3D.
- Alex Kidd in Miracle World DX and Dreaming Sarah now boot and render correctly (previously failed to start or rendered corrupted).
- Not yet verified for regressions on NVIDIA hardware; testing from someone with an NVIDIA GPU would help confirm no regression there.

# AI assistance disclosure
AI assistance was used to diagnose the root cause of each issue and prepare the initial patch. The contributor reviewed the complete diff and tested it on the affected AMD hardware before opening this pull request.

### DIFF
--- a/src/graphics/host_gpu/renderer/shaderSubgroup.cpp
+++ b/src/graphics/host_gpu/renderer/shaderSubgroup.cpp
@@ -6,9 +6,14 @@ namespace Libs::Graphics {
 
 ShaderLaneMaskMode SelectGraphicsLaneMaskMode(const GraphicContext& context,
                                               uint32_t              guest_wave_size) {
-	return context.subgroup_size == 32u && guest_wave_size == 64u
-	           ? ShaderLaneMaskMode::PerInvocation
-	           : ShaderLaneMaskMode::NativeWave;
+	// A 64-wide guest wave can't assume the host GPU's lane layout matches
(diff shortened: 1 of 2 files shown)


ahmeddzairdev-ctrl:
are you willing to build it ? i wanna test it on my rx 7600

Nmzik:
@kirathenotebook Thank you! I’ll test it on my NVIDIA GPU and report back on whether it still works

PR #32, opened 2026-07-17
Enhance memory management and logging for system modules

Pouare514:
- Implement `RegisterSystemModuleMemory` for improved memory registration of system modules.
- Add `SumBytesOfType` to track memory usage by virtual range types.
- Update logging to provide detailed insights into memory types and sizes during queries.
- Refactor memory range checks to accommodate both program and system module types for accurate management.


Pouare514:
rwaaar i will wait a bit then cause i will always have the previous commits, i would need to retake the base code to do it clean so idk how to do it tbh (without having to stop pushing code for some days)

Nmzik:
> rwaaar i will wait a bit then cause i will always have the previous commits, i would need to retake the base code to do it clean so idk how to do it tbh (without having to stop pushing code for some days)

All good. No worries

PR #36, opened 2026-07-18
audio: rely on SDL queue pacing

Digote:
## Summary

- remove the legacy wall-clock pacing from `AudioOutOutputs`
- use the actual SDL audio queue as the pacing source
- preserve the existing queue-depth backpressure for blocking output
- handle null buffers as blocking drain requests

## Problem

The wall-clock delay was inherited from the original Kyty implementation, where
audio output was not implemented and the buffer duration was only simulated.

KytyPS5 now sends real audio through `SDL_QueueAudio`. The existing code already
applies backpressure based on the amount of audio queued in SDL. Keeping the
legacy delay in `AudioOutOutputs` introduced a second pacing mechanism before
queueing each buffer.

With 256-frame buffers at 48 kHz, each buffer represents approximately 5.33 ms
of audio. Sleeping for that duration before queueing the next buffer prevented
SDL from consistently maintaining its intended buffering cushion, making
underruns and audible stuttering more likely under scheduler or emulation
jitter.

## Implementation

The per-port `last_output_time` state and the corresponding wall-clock sleep
were removed.

The existing backpressure in `QueueSdlAudio` is preserved. In blocking mode, the
function still waits while SDL has more than two buffers queued before adding
the next buffer. Therefore, removing the wall-clock delay does not allow the
normal audio queue to grow without bounds.

Non-blocking output behavior is unchanged and continues to skip the SDL queue
wait.

A null buffer is handled separately as a blocking drain request. It waits for
the actual SDL queue to finish, with the existing 200 ms watchdog behavior to
avoid an indefinite wait if the audio device stops consuming data.

## Validation

- Windows Release build with `clang-cl` completed successfully
- installation completed successfully
- `kyty_emulator.exe --help` smoke test passed
- tested Dreaming Sarah (`PPSA02929`) with 256-frame FloatStereo audio at 48 kHz
- audio playback was noticeably smoother after the change
- the new game log contains no AudioOut or SDL audio errors
- no automated AudioOut tests are currently configured

## AI assistance disclosure

AI assistance was used to inspect the inherited timing implementation, analyze
the SDL queue behavior, review edge cases, and prepare the initial change.

The contributor manually tested the affected game, confirmed the audio
improvement, and reviewed the final one-file diff before submission.


red-prig:
From shadps4's experience, I can say that this is incorrect. Time delays should always be simulated manually because games expect them, and SDL doesn't guarantee the same time delays in each implementation, so you'll get bugs from some people and normal behavior from others.

Digote:
Thank you very much for the detailed explanation. I'll close this PR and work on a new one that better contributes to the emulator.

PR #89, opened 2026-07-20
loader: narrow PLT0 AV patch (M1W2 v1.2 + v1.3)

abdulazizFarhan:
# loader: narrow PLT0 AV patch (M1W2 v1.2 + v1.3)

## Problem

PS5 ELF binaries use a 0x1/0x0 sentinel value in their PLT0 slot to mark
**unresolved** function imports. When the binary later executes a `call
[rax]` / `call [rbx]` whose target is the placeholder, the CPU raises an
Execute access-violation (`0xC0000005` / `av_type=Execute, av_addr=0x0|0x1`).
kyty's existing exception handler has no fallback for this case and calls
`EXIT(...)` → fail-fast → emulator dies. This blocked ~15 of 30 PS5 ELFs
from reaching the Vulkan robustness gate (baseline ~25/30 before patch).

A second, downstream class of AV was also visible: after the v1.2 patch
unblocked the call site, the next instruction often hit
`av_type=Write, av_addr=0x0` (a `mov [rax], ...` whose source was the
stubbed function's return value, now a NULL pointer in `rax`).

## Fix

Single new block in `KytyExceptionHandler` (`runtimeLinker.cpp:1017-1110`):

1. **Trigger**: `av_type=Execute` OR `av_type=Write` AND
   `0 < av_vaddr < 0x100000` AND `av_vaddr != g_invalid_memory (0x84000000)`.
2. **Resolve `fault_ip`**: prefer `ctx->Rip` from the saved Windows CONTEXT
   (which is more reliable than `ExceptionAddress` for these PS5 faults,
   where the kernel often reports the access addr as `Rip`). Fall back to
   `info->exception_address`, then to `SysStackWalkX86(stack[0] - 2)`.
3. **Patch**: 32 bytes (one cache line) at `fault_ip & ~0x1F` with
   `0x90` (NOP). 32 bytes is enough to cover any single call/jmp
   (max 7 bytes on x64) without clobbering adjacent valid code (128
   bytes overwrote the next instruction and triggered IllegalInstruction
   on resume — verified in V18b).
4. **Resume**: `ctx->Rip = fault_ip + 16` (skip past the bad call/jmp);
   `ctx->Rax = 0` (well-defined "function not found" return so the next
   instruction that consumes the return value doesn't read garbage).
5. **Tracking**: `static std::unordered_set<uint64_t> patched_bases` so
   each call site gets patched exactly once.

Linux path is a no-op for the patch logic — the host exception layer
there already supports setting the new RIP via `ucontext_t`.

## Sweep

V22 — 30 PS5 ELFs, 5s/elf timeout:

| Metric | Baseline (V19) | v1.2 (V20) | v1.3 (V22) |
|---|---|---|---|
| Vulkan robustness gate | 25/30 | 24/30 | 21/30 |
| M1W2 patch fires | – | 14/30 | 18/30 |
| Total `Relocate: unresolved` patches | 1709 | 1845 | – |
| Fatal / fail-fast | 0 | 0 | 0 |

v1.3 down from 5 remaining Write AVs (V21) to 1 (V22). The remaining
`hello_argv: Access violation: Write [0xffffffffffffff85]` is a kernel
no-access area write — different class, not PLT0.

**Known regression**: `elfldr.elf` regressed from V19 baseline because
kyty's GPU `HandleFault` consumes the AV before the M1W2 block runs.
This is structural (the order is `HandleFault` → our block) and out of
scope for this PR. Documented in the plan.

## Files

- `src/loader/runtimeLinker.cpp` — 1 file, +94 / -0, all additive in a
  single new block inside `KytyExceptionHandler`. No new public API,
  no header changes.

## Root cause writeup (long form)

`exception_record->ExceptionAddress` is 0 for these PS5 binary AVs on
Windows because the kernel reports the access addr (`0x0` or `0x1`) as
`Rip` in some vectored-exception code paths. The actual faulting
instruction is in the saved `CONTEXT.Rip` field. Before v1.2, kyty
only had `info.exception_address` available; v1.2 reads
`info.native_context` (the saved CONTEXT) to get the real RIP. Where
`ctx->Rip` is also bogus (< 0x10000), the handler falls back to walking
the call stack with `SysStackWalkX86` and uses `stack[0] - 2` as the
call-site approximation (subtract 2 to back up past the `call` opcode
or the `jmp` indirect prefix).

The 32-byte patch window is the smallest unit that aligns to a cache
line and covers the longest possible call/jmp encoding on x64. NOPs
across the window are safe because the instruction that produced the
AV is a single `call [reg]` (3 bytes) or `jmp [reg]` (3 bytes) — the
rest of the cache line is unrelated code that will simply execute
past without effect.

The `Rax = 0` resume value is chosen because callers typically use
the return as a NULL check; returning 0 ("function not found") keeps
the path deterministic without requiring per-call stubbing.

## AI-assisted disclosure

This change was developed with AI assistance (Claude/oh-my-claudecode).
All edits and sweep runs were performed locally with the user's 30-ELF
test corpus.

## Test/verify

```bash
# build
cmake --build _Build/windows --target kyty_emulator.exe

# 30-ELF sweep (5s timeout each, full results in .omc/state/cross-game-runs-v22/)
for f in .omc/test-games/discovered/*.elf; do
  _Build/windows/kyty_emulator.exe --game "$f" 2>&1 | tail -1
done
```

### DIFF
--- a/src/emulator.cpp
+++ b/src/emulator.cpp
@@ -189,6 +189,10 @@ void Run(const RunOptions& options) {
 		EXIT("ELF is required\n");
 	}
 
+	// M1W3: enable the headless FPS stdout summary if --show-fps was
+	// set. Must happen before WindowRun() spawns the rendering thread.
+	Libs::Graphics::SetFpsStdoutEnabled(options.show_fps);
+
 	Init(options.config);
 
 	ClearDebugTextureFolder();
--- a/src/emulator.h
+++ b/src/emulator.h
@@ -12,6 +12,11 @@ struct RunOptions {
 	Config::ConfigOptions config;
 	std::filesystem::path app0_dir;
 	std::filesystem::path elf;
+	// M1W3: when true, the rendering thread prints a once-per-second
+	// FPS summary to stdout (e.g. "[FPS] 23.4 frames/sec (t=15.0s)").
+	// Independent of any in-window text overlay; intended for headless
+	// measurement sessions.
+	bool show_fps = false;
 };
 
 void Run(const RunOptions& options);
--- a/src/graphics/host_gpu/renderer/depthRenderTarget.cpp
+++ b/src/graphics/host_gpu/renderer/depthRenderTarget.cpp
@@ -238,8 +238,8 @@ void ResolveRenderDepthTarget(uint64_t submit_id, CommandBuffer* buffer, const H
 	r->format =
 	    ResolveHostDepthAttachmentFormat(g_render_ctx->GetGraphicCtx(), *policy, has_stencil);
 	if (r->format == VK_FORMAT_UNDEFINED) {
-		DepthFatal("no host depth/stencil format supports required usage for %s",
-		           string_VkFormat(ideal_format));
+		const auto fmt_str = string_VkFormat(ideal_format);
+		DepthFatal("no host depth/stencil format supports required usage for %s", fmt_str.c_str());
 	}
 	const uint32_t guest_format = Prospero::GpuEnumValue(policy->guest_format);
 	const uint32_t bytes        = policy->bytes_per_element;
--- a/src/graphics/presentation/window.h
+++ b/src/graphics/presentation/window.h
@@ -25,6 +25,12 @@ PreparedFrame* WindowPrepareBlankFrame(CommandBuffer* buffer, uint32_t width, ui
                                        bool opaque);
 void           WindowPresentFrame(PreparedFrame* frame);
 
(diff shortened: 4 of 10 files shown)


abdulazizFarhan:
Superseded by #91 (M1W2 v1.4), which reverts the v1.3 Write-AV trigger and RAX=0 set that regressed Vulkan robustness from 24/30 to 21/30. v1.4 restores 30/30 reach.

PR #97, opened 2026-07-22
Add graphics regression testing

Nmzik:
## Summary

- add exact presented-frame baseline recording and comparison with enforced environment provenance
- add versioned commands-only GPU traces with checksummed completion footers
- add isolated single-game and sequential suite runners with process-tree timeouts
- add focused native/Python tests, documentation, and a trusted self-hosted GPU workflow

## Reliability

- atomic, durable manifests and frame artifacts with cross-process output reservations
- strict schema, bounds, corruption, provenance, baseline-anchor, and output-collision validation
- per-case work directories, logs, reports, summaries, and retained failure evidence
- suspended Windows process launch before Job Object assignment
- explicit artifact allowlist; frame evidence is opt-in and work directories are never uploaded
- commercial game content and baselines remain outside the repository

## Validation

- GPU trace tests
- frame regression tests
- 18 graphics regression runner tests
- CLI smoke checks
- C++20 no-exceptions warning-as-error focused compilation
- production source syntax checks
- three independent final audits with no release blocker found

### DIFF
--- a/.github/workflows/graphics-regression.yml
+++ b/.github/workflows/graphics-regression.yml
@@ -1,0 +1,87 @@
+name: Graphics Regression Suite
+
+on:
+  workflow_dispatch:
+    inputs:
+      suite_path:
+        description: Absolute path to the private suite file on the runner
+        required: true
+        type: string
+      emulator_path:
+        description: Absolute path to kyty_emulator.exe on the runner
+        required: true
+        type: string
+      artifact_name:
+        description: Uploaded result artifact name
+        required: true
+        default: graphics-regression-results
+        type: string
+      allow_environment_mismatch:
+        description: Deliberately override baseline environment checks
+        required: true
+        default: false
+        type: boolean
+      upload_frame_evidence:
+        description: Upload raw, preview, and difference frames (may contain game imagery)
+        required: true
+        default: false
+        type: boolean
+
+jobs:
+  regression:
+    runs-on: [self-hosted, Windows, X64, gpu]
+    environment: graphics-regression
+    timeout-minutes: 360
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
+
+      - name: Run serialized suite
+        shell: pwsh
+        env:
+          KYTY_SUITE: ${{ inputs.suite_path }}
+          KYTY_EMULATOR: ${{ inputs.emulator_path }}
+          KYTY_OUTPUT: ${{ runner.temp }}\kyty-graphics-regression-${{ github.run_id }}-${{ github.run_attempt }}
+          KYTY_ALLOW_ENVIRONMENT_MISMATCH: ${{ inputs.allow_environment_mismatch }}
+        run: |
+          New-Item -ItemType Directory -Force -Path $env:KYTY_OUTPUT | Out-Null
+          $arguments = @(
+            "tools\run_graphics_regression_suite.py",
+            "--suite", $env:KYTY_SUITE,
+            "--emulator", $env:KYTY_EMULATOR,
+            "--output", $env:KYTY_OUTPUT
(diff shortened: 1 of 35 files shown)


Nmzik:
@codex review

src/graphics/regression/frameRegressionVulkan.cpp
```diff
@@ -0,0 +1,181 @@
...
+		const auto row_pitch = image.extent.width * bytes_per_pixel;
+		const auto byte_size = static_cast<uint64_t>(row_pitch) * image.extent.height;
+		if (byte_size > std::numeric_limits<size_t>::max() || byte_size > MaxFrameBytes) {
+			Fail("readback", "presented frame is too large for host memory");
+			return true;
+		}
+
+		std::vector<uint8_t> bytes(static_cast<size_t>(byte_size));
+		// Vulkan bufferRowLength is measured in texels; FrameView::row_pitch is measured in bytes.
+		Transfer::DownloadImage(bytes.data(), byte_size, image.extent.width, image,
```
Nmzik, line 98: Synchronize the producing graphics submission before this utility-queue readback. `frame.image` is filled by the flip-preparation command buffer; for CPU flips, the request becomes ready immediately after `m_scheduler.Flush()`, not after its fence completes. `DownloadImage` then submits a separate immediate command on `QUEUE_UTIL`, and this backend has no cross-queue scheduler or semaphore at this boundary. Under GPU load the readback can overtake the prepared-frame copy and hash stale or nondeterministic contents. Wait on the preparation fence/semaphore or record the readback on the same graphics queue.

Nmzik (commented):
Blocking review findings: the current implementation can produce a regression result from the wrong frame data or the wrong executable revision. Please address the synchronization and revision-binding findings before treating this as a trustworthy regression gate. The previously reported generated-version dependency issue also remains applicable.

.github/workflows/graphics-regression.yml
```diff
@@ -0,0 +1,87 @@
...
+          KYTY_SUITE: ${{ inputs.suite_path }}
+          KYTY_EMULATOR: ${{ inputs.emulator_path }}
+          KYTY_OUTPUT: ${{ runner.temp }}\kyty-graphics-regression-${{ github.run_id }}-${{ github.run_attempt }}
+          KYTY_ALLOW_ENVIRONMENT_MISMATCH: ${{ inputs.allow_environment_mismatch }}
+        run: |
+          New-Item -ItemType Directory -Force -Path $env:KYTY_OUTPUT | Out-Null
+          $arguments = @(
+            "tools\run_graphics_regression_suite.py",
+            "--suite", $env:KYTY_SUITE,
+            "--emulator", $env:KYTY_EMULATOR,
```
Nmzik, line 56: Bind the tested executable to the checked-out revision. This job checks out the selected ref but executes an arbitrary pre-existing `emulator_path`; it neither builds that checkout nor verifies the executable build ID against `github.sha`. Because regression environment compatibility intentionally excludes `build_id`, a stale binary can produce a passing run for the wrong revision. Build the emulator in this job, or pass an expected revision into the runner and reject a mismatched run provenance.

src/common/platform/sysLinuxFileIO.cpp
```diff
@@ -363,7 +363,7 @@ bool SysFileDeleteFile(const std::filesystem::path& name) {
 
 bool SysFileFlush(sys_file_t& f) {
 	if (f.type == SYS_FILE_FILE && f.f != nullptr) {
-		return (fflush(f.f) == 0);
+		return fflush(f.f) == 0 && fsync(fileno(f.f)) == 0;
```
Nmzik, line 366: Avoid changing the meaning of every existing `Common::File::Flush()` call to a durable disk sync. `TraceWriter::Record()` flushes after every GPU event, so this turns tracing into an `fsync` per submission on POSIX, and unrelated existing callers also inherit the cost. Keep ordinary flush behavior and add a separate durable-sync operation for finalization/atomic artifacts, or batch trace durability at an explicit boundary.

PR #101, opened 2026-07-22
fix Windows fixed placeholder reservations below allocation granularity

### LINKED ISSUE #100: [GAME BUG]: UFC 5 (DOES NOTHING)
### Game title

UFC 5

### Game ID / serial

PPSA03541

### KytyPS5 version

dbcb444

_No response_

### What happened?

Game crashes on launch.

### Steps to reproduce

Open emulator.
Boot game.

### Expected behavior

Splash screens would render, then game menu.

### OS

Windows 11 25H2

### CPU

AMD Ryzen 5 5600

### GPU

Nvidia GeForce RTX 4060 TI

### RAM / VRAM

32 GB DDR4 RAM / 8 GB VRAM

### Log file upload

[PSA03541_kyty.txt](https://github.com/user-attachments/files/30276368/PSA03541_kyty.txt)

### Screenshots/video

Image

### Extra notes

_No response_

Digote:
## What changed

- Fix Windows fixed placeholder reservations when a PS5 16 KiB mapping starts at an address that is not aligned to the Windows allocation granularity.
- Reserve the enclosing allocation-granularity placeholder container, split it, release the unused neighboring ranges, and retain an exact placeholder for the requested range.
- Add `FixedSub64KDirectMapUsesExactAddress`, covering fixed reservation, exact direct-memory mapping, shared-backing aliasing, unmap, placeholder restoration, and release of the neighboring host ranges.

## Why

UFC 5 fails early in #100 while mapping `EAOSGlobal` at `0xfbffffc000` with a size of `0x4000`.

`VirtualAlloc2` requires a non-null fixed base address to be aligned to the system allocation granularity. The affected address is a valid 16 KiB PS5-page address but is offset within a 64 KiB Windows allocation-granularity range.

The fix reserves an aligned placeholder container and uses placeholder splitting so that `MapViewOfFile3(..., MEM_REPLACE_PLACEHOLDER)` receives a placeholder whose base address and size exactly match the requested mapping.

## Documentation

- https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualalloc2
- https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-mapviewoffile3
- https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualfree

## Validation

- Built on Windows with the project CMake/Ninja configuration:
  `cmake --build Build\windows --target kyty_emulator virtual_memory_allocation_tests`
- Ran `virtual_memory_allocation_tests`: 27/27 cases passed.
- `git diff --check` passed.
- Ran clang-format validation for `src/kernel/memoryAddressSpace.inc`.
- Reproduced UFC 5 (`PPSA03541`): the original `placeholder-reserve-failed` / fatal error no longer occurs. `EAOSGlobal` now maps at `0xfbffffc000` with `shared = yes`.

## Scope

This partially addresses #100. The new UFC 5 log proceeds far beyond the original memory failure but later stops on an unrelated `unsupported storage texture` renderer error, so this PR should not close #100.

## AI assistance disclosure

AI assistance was used for investigation, implementation support, test design, and documentation research. I have reviewed the final diff and validation results and remain responsible for this contribution.


src/kernel/memoryAddressSpace.inc
```diff
@@ -1119,20 +1119,47 @@ public:
...
+			return false;
+		}
+
+		const auto left_size  = vaddr - container_start;
+		const auto right_size = container_end - (vaddr + size);
+		if (left_size != 0) {
+			ReleaseFree(container_start, left_size);
+		}
+		if (right_size != 0) {
+			ReleaseFree(vaddr + size, right_size);
```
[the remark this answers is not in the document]
Digote: Thanks, good catch! 👍

I fixed this in a5637b3.

The old code was releasing the neighboring placeholder ranges too early. Now it keeps the full 64 KiB placeholder reserved, so another adjacent 16 KiB reservation can still use the same block.

I also added a regression test for that adjacent 16 KiB case. All Windows virtual-memory tests are passing (27/27).

src/kernel/memoryAddressSpace.inc
```diff
@@ -1119,21 +1119,35 @@ public:
...
+		                          container_size, MEM_RESERVE | MEM_RESERVE_PLACEHOLDER,
+		                          PAGE_NOACCESS, nullptr, 0);
+		if (ptr == nullptr || reinterpret_cast<uint64_t>(ptr) != container_start) {
 			if (ptr != nullptr) {
 				VirtualFree(ptr, 0, MEM_RELEASE);
 			}
 			return false;
 		}
-		AddFree(vaddr, size);
+		AddFree(container_start, container_size);
```
Nmzik: This retains the entire aligned container in `m_free`, but the virtual-range layer records only the requested 16 KiB range. When that range is later released, `ReleaseFree(vaddr, size)` frees only those 16 KiB; the unowned sibling placeholders remain reserved and tracked for the rest of the process. Repeating isolated sub-granularity reservations therefore leaks up to 48 KiB of host address space per 16 KiB guest range and can block unrelated host reservations in those gaps. Please release the enclosing container once none of its subranges are owned, and extend the test to verify all sibling addresses become unmapped after the requested ranges are released.

Nmzik (commented):
The adjacent-page fix addresses the earlier review finding, but there is one remaining non-blocking resource-lifetime issue in the retained sibling placeholders. CI could not be inspected because the workflow stopped before creating any jobs.

Digote:
Thanks for the detailed explanation. I've addressed the container lifetime issue in 399dc7f.

Fixed allocation-granularity containers are now tracked separately from their individual guest ranges. When a range gets released, we check the virtual-range ownership of each 64 KiB container:

- If another guest subrange still occupies the container, all sibling placeholders are kept.
- Once the last owner is removed, every retained placeholder fragment in the container is released and the container gets pulled from tracking.

Same cleanup applies to the unmap and rollback paths too, so a failed mapping can't leave an unowned container behind.

I extended FixedSub64KDirectMapUsesExactAddress to check that:
- releasing the first 16 KiB range keeps its placeholder while the adjacent range is still owned
- after releasing the final owner, all four 16 KiB addresses in the 64 KiB container report MEM_FREE
- the full 64 KiB container can be reserved again at the same address and released cleanly

Validation:
- cmake --build Build\windows --target virtual_memory_allocation_tests
- virtual_memory_allocation_tests: 27/27 passing
- git diff --check clean
- clang-format clean

Also reproduced UFC 5 again — EAOSGlobal still maps fine at 0xfbffffc000 with shared backing, no placeholder reservation or container-release errors.

src/kernel/memoryAddressSpace.inc
```diff
@@ -1119,21 +1119,44 @@ public:
...
+		// VirtualAlloc2 requires a fixed base address to use the host allocation
+		// granularity. Reserve and retain the enclosing container as placeholders;
+		// Consume() will isolate the exact PS5-page range for MapViewOfFile3.
+		const auto container_start = vaddr & ~(m_granularity - 1u);
+		if (UINT64_MAX - (m_granularity - 1u) < vaddr + size) {
+			return false;
+		}
+		const auto container_end  = AlignUp(vaddr + size, m_granularity);
+		const auto container_size = container_end - container_start;
+		auto* ptr = VirtualAlloc2(GetCurrentProcess(), reinterpret_cast<void*>(container_start),
```
Nmzik: `ReserveFixed` returns early only when one `m_free` entry contains the entire request. For example, after reserving `[base, base + 0x4000)`, a valid non-overlapping reserve of `[base + 0x4000, base + 0x14000)` spans the remainder of that tracked 64 KiB container and the first page of the next container. `FindContaining` fails, then this call tries to reserve both aligned containers at once; Windows rejects it because the first container is already reserved, so the guest request incorrectly returns `ENOMEM`. Please combine existing tracked placeholder containers with reservations for only the missing containers, and add a regression covering a cross-container request with the first container already present.

Nmzik (changes requested):
The new container-lifetime cleanup addresses the previously reported sibling-placeholder leak, but valid fixed reservations can still fail when they cross from an existing tracked 64 KiB container into an unreserved one. CI could not be inspected because the workflow stopped before creating any jobs.

Digote:
Thanks, fixed in 2c41030. ReserveFixed now reuses tracked 64 KiB containers, reserves only missing ones, and rolls back newly created containers on failure. I added the exact cross-container regression case; all 28 Windows virtual-memory tests pass, and UFC 5 still maps EAOSGlobal correctly without placeholder errors.

Digote:
It looks like the UFC 5 issue has already been resolved with https://github.com/KytyPS5/KytyPS5/pull/135
Closing.

Nmzik:
@Digote Didnt expect that change to fix UFC5 memory allocation bug. Thanks for reporting

Digote:
@Nmzik test https://github.com/KytyPS5/KytyPS5/commit/d8a4c83cc7d343b1c7a2702084f3c282fbd00cd4
Teste

PR #109, opened 2026-07-26
New QML launcher + host input remapping

DistantMyth:
New Qt6/QML launcher (dark theme, sidebar nav, cover art, blurred backdrops) inspired by SharpEmu's UI, plus full keyboard/mouse → DualSense remapping with a graphical controller layout. Fixes the host-key fallback so remapped keys don't double-fire with the defaults.


DistantMyth:
I know i took a bit of time but here it is
image

Crispy81:
Looks good from the screenshot. Keen to see it in action. Nice work.

src/launcher/qml/views/EmulationView.qml
```diff
@@ -0,0 +1,260 @@
...
+                            Column {
+                                Layout.fillWidth: true
+                                spacing: 2
+                                Text { text: "Unified GDDR6 Memory Limit"; font.pixelSize: 14; font.weight: Font.DemiBold; color: window.textPrimary }
+                                Text { text: "Reserve RAM limit for guest system heap and VRAM allocation."; font.pixelSize: 11; color: window.textDim }
+                            }
+
+                            ComboBox {
+                                id: memCombo
+                                model: ["16 GB (Full Native)", "12 GB (Optimized)", "8 GB (Minimum)"]
```
Nmzik: What is it? We’re only reserving address space from the OS, not allocating memory immediately, so this doesn’t make any sense.

Nmzik:
@DistantMyth Thank you for the PR. Let’s keep it simple for now: just the new UI and remapping. Everything related to audio and CPU threads/SMT can be handled in separate PRs. There’s too much questionable stuff here. Thanks for understanding.

src/launcher/qml/views/EnvironmentView.qml
```diff
@@ -0,0 +1,240 @@
...
+                ToggleRow {
+                    titleText: "Reload Shaders on Focus"
+                    descText: "Recompile shaders when the emulator window regains focus (helps shader iteration)."
+                    checked: launcherBridge.reloadShadersOnFocus
+                    onToggled: function(v) { launcherBridge.reloadShadersOnFocus = v }
+                }
+
+                ToggleRow {
+                    titleText: "Async Shader Compilation"
+                    descText: "Compile shaders asynchronously to avoid stuttering (may produce visual glitches briefly)."
```
Nmzik: Please remove it for now

DistantMyth: Yeah definately

src/launcher/qml/views/EmulationView.qml
DistantMyth: i see, it was kindof rushed, removing it.

DistantMyth:
I still left the icons there if the things are to be implemented in near future, i can remove them if you want.
Also another thing which i failed to mention last time was that i made a logo for the project, do you want me to keep it or you prefer otherwise.
I took care of the other things, just added input remapping and redesigned the UI, nothing else, just as you asked for.

src/launcher/src/launcher_qml.cpp
```diff
@@ -0,0 +1,693 @@
...
+	if (m_process->state() == QProcess::Running) {
+		qWarning() << "[KytyPS5 Launcher] Emulator is already running!";
+		return;
+	}
+
+	QDir search_dir(QApplication::applicationDirPath());
+#ifdef __linux__
+	QString exeName = "kyty_emulator";
+#else
+	QString exeName = "kyty_emulator.exe";
```
Nmzik, line 413: [blocking] This treats every non-Linux host as Windows. The macOS build installs an extensionless `kyty_emulator`, so the default QML launcher always searches for `kyty_emulator.exe` and refuses to launch any game there. Use a Windows-specific conditional for the `.exe` suffix and the extensionless name on Linux/macOS.

src/launcher/src/launcher_qml.cpp
```diff
@@ -0,0 +1,693 @@
...
+	QString gamePath = path;
+	QFileInfo fi(path);
+	if (fi.isDir()) {
+		gamePath = QDir(path).filePath("eboot.bin");
+		if (!QFile::exists(gamePath)) {
+			gamePath = path; // Fallback to dir
+		}
+	}
+
+	args << "--game" << gamePath;
```
Nmzik, line 481: [blocking] The replacement launch path drops the legacy launcher's automatic `--game-patch` argument (`PatchesDialog::PatchPlanPath(info.title_id)`). Any saved patch plan is silently ignored when launching from the new default UI. Carry the selected title ID into this path and append `--game-patch` when its plan exists.

src/launcher/src/launcher_qml.cpp
```diff
@@ -0,0 +1,693 @@
...
+	// contains eboot.bin as a game. Matches the behaviour of the legacy Qt
+	// Widgets launcher (ConfigurationListWidget::ScanGameDirectory). Without
+	// this, adding a folder like "D:\Extra" (which holds many game subfolders)
+	// shows just one bogus "Extra" entry instead of each game inside it.
+	for (const QString& root_path : std::as_const(m_savedGameDirs)) {
+		QDir root(root_path);
+		if (root_path.isEmpty() || !root.exists()) continue;
+
+		QList<QDir> pending;
+		const auto  subdirs =
```
Nmzik: [blocking] The scan begins with the selected root's children, so choosing a game directory that itself contains `eboot.bin` never examines or adds that game. This conflicts with the UI's “Add Game Folder” action. Queue the root itself first, then descend only when it is not a game.

src/graphics/presentation/window/window.cpp
```diff
@@ -273,6 +328,20 @@ static void GameEventKeyboard(WindowLoopState& game, const EventKeyboard& key) {
...
 
+static uint32_t MouseToPadButton(const std::string& mouse_event) {
+	// mouse_event is e.g. "Left", "Right", "Middle".
+	const auto& keymap = Config::GetKeymap();
+	for (const auto& kv : keymap) {
+		const std::string& pad_name = kv.first;
+		const std::string& binding  = kv.second;
+		if (binding.rfind("Mouse:", 0) != 0) continue;
+		if (binding.substr(6) == mouse_event) {
+			return PadButtonFromName(pad_name);
```
Nmzik, line 338: [blocking] The mouse bindings seeded by `defaultKeyBindings()` cannot resolve here: `TouchClick`, `MouseTouch`, `TouchpadX`, and `TouchpadY` are not recognized by `PadButtonFromName`, and motion events are excluded by `GameEventMouse`. Consequently the advertised mouse remapping emits no controller event with the supplied defaults. Map these bindings to implemented controller semantics (including motion if exposed), or remove the unsupported mappings.

src/launcher/qml/views/InputMappingView.qml
```diff
@@ -0,0 +1,520 @@
...
+                                                id: clearMouse
+                                                anchors.fill: parent
+                                                hoverEnabled: true
+                                                cursorShape: Qt.PointingHandCursor
+                                                onClicked: launcherBridge.setBinding(modelData.pad, "")
+                                            }
+                                        }
+                                    }
+
+                                    MouseArea {
```
Nmzik: [blocking] This full-row `MouseArea` is declared after the row content, so it is stacked above `setMouse` and `clearMouse` and consumes their clicks. The Set/Clear controls cannot perform their own actions. Put the row handler behind the controls or explicitly avoid/propagate events over those button regions.

Nmzik (changes requested):
Blocking launcher and input regressions remain. The existing request to keep this PR limited to the UI and remapping also remains applicable: audio changes are still present in the current head.

DistantMyth:
Updated

Nmzik:
Host input remapping has been merged along with multiple fixes. That's all we need for now

PR #111, opened 2026-07-27
launcher: add local game patches

Nmzik:


src/loader/gamePatch.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+		if (add_read &&
+		    !Common::VirtualMemory::Protect(segment_address, segment.p_memsz, old_mode)) {
+			return Fail(error, "could not restore executable segment protection");
+		}
+	}
+
+	::printf("Game patch: found %zu entries for '%s'\n", match_count,
+	         write->patch_name.c_str());
+	if (match == 0) {
```
Nmzik, line 214: Blocking: reject ambiguous signatures instead of patching the first match. `match_count` already records every occurrence, but this condition only rejects zero matches. A short or non-unique byte sequence can therefore select whichever occurrence happens to be scanned first and modify unrelated guest code. Require exactly one match, or include an explicit disambiguating offset/range in the plan.

src/loader/gamePatch.cpp
```diff
@@ -0,0 +1,265 @@
...
+	for (auto& write: plan->writes) {
+		if (!ResolveWrite(program, &write, error)) {
+			return false;
+		}
+	}
+	return true;
+}
+
+bool ApplyWrites(Plan* plan, std::string* error) {
+	for (auto& write: plan->writes) {
```
Nmzik, line 231: Blocking: validate the resolved write set for overlapping ranges before applying it. Every enabled write is resolved against the original image, then these writes are committed sequentially without rechecking `expected`; two patches targeting the same or overlapping bytes both pass preparation and the later write silently overwrites the earlier one. Reject conflicts (or define and validate explicit composition semantics) before changing memory.

Nmzik (commented):
Blocking findings remain in the patch application path. The current implementation can select an unintended guest address when a signature is ambiguous and can silently compose conflicting enabled patches. Please make target resolution unambiguous, reject overlapping ranges before mutation, and add focused tests for zero/one/multiple matches and conflicting writes.

PR #102, opened 2026-07-23
macOS (Apple Silicon) support: native build running under Rosetta 2 with MoltenVK

m1abdullahh:
## Overview

This adds a native macOS build target for Apple Silicon. The emulator compiles and runs as an x86_64 binary under Rosetta 2 (guest PS5 code executes natively as x86_64; the emulator itself is translated), with graphics going through Vulkan to MoltenVK to Metal. It builds on the current tree, boots a commercial title into gameplay, and responds to controller input.

## What was tested

Raiden III x MIKADO MANIAX boots through the loader, Vulkan init, swapchain, and window creation into guest execution, renders frames continuously, and takes controller input; the audio pipeline is wired up. It is a functional port with a couple of known rendering artifacts and performance still to improve (see Known issues), not a finished product, but well past proof of concept.

Verified locally on an M3 Pro MacBook Pro: clean configure, build, link, JIT codesign, and launch under Rosetta 2.

## Building on macOS

Prerequisites: Apple Silicon Mac, Rosetta 2, Xcode Command Line Tools, CMake, Ninja, and a Vulkan runtime that provides MoltenVK (for example the LunarG Vulkan SDK for macOS).

```sh
git clone --recursive https://github.com/<you>/KytyPS5
cd KytyPS5
cmake -S src -B _Build/macos -G Ninja \
  -DCMAKE_BUILD_TYPE=Release \
  -DCMAKE_OSX_ARCHITECTURES=x86_64
cmake --build _Build/macos --target kyty_emulator --parallel
```

The build produces an x86_64 binary and codesigns it with the JIT entitlements automatically. Run a game with `--game <dir|elf>` (see `--help` for options). If the window fails to appear because of a Rosetta issue in the macOS window chrome, set `KYTY_BORDERLESS=1` for a borderless window.

## Known issues

- Some rendering artifacts remain: video noise during cutscenes and a black box around certain sprites. These are on the texture-upload side and under investigation; they do not block boot or input.
- Frame rate is modest (around 26 fps in the tested title). This appears consistent with the title's general performance and is not macOS-specific.
- The guest Ngs2 audio path uses a stub for one call, which may leave a gap in some in-game audio.

PR #114, opened 2026-07-27
ci: add macOS build workflow

Nmzik:
## Summary

- add a macOS 15 GitHub Actions build for the launcher
- select Xcode 26 and target x86_64 with Qt 6.10.3, Ninja, and glslang
- verify the installed binaries and emulator signature
- upload the macOS artifact and add a README status badge
- include the minimal macOS POSIX compatibility guards from #102

## Why

The project has macOS x86_64 build support but no dedicated CI job to exercise it.

## Validation

- Actionlint
- workflow structure parsing
- `git diff --check`
- GitHub Actions macOS build

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -74,8 +74,8 @@ jobs:
         run: |
           file _Build/macos/install/launcher
           file _Build/macos/install/kyty_emulator
-          lipo -verify_arch x86_64 _Build/macos/install/launcher
-          lipo -verify_arch x86_64 _Build/macos/install/kyty_emulator
+          lipo _Build/macos/install/launcher -verify_arch x86_64
+          lipo _Build/macos/install/kyty_emulator -verify_arch x86_64
           codesign --verify --strict _Build/macos/install/kyty_emulator
 
       - name: Upload artifacts
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -1,0 +1,78 @@
+name: Build KytyPS5 (macOS)
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [ main, master ]
+  pull_request:
+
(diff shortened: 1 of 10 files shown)


Nmzik (commented):
Blocking findings are included inline. The macOS workflow currently fails before install or artifact verification; the Windows and Linux checks are still running.

.github/workflows/build-macos.yml
```diff
@@ -0,0 +1,86 @@
...
+          file _Build/macos/install/kyty_emulator
+          lipo _Build/macos/install/launcher -verify_arch x86_64
+          lipo _Build/macos/install/kyty_emulator -verify_arch x86_64
+          codesign --verify --strict _Build/macos/install/kyty_emulator
+
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: KytyPS5-macOS-x86_64
+          path: _Build/macos/install/**
```
Nmzik, line 85: Blocking: this uploads only the raw `cmake --install` directory. `src/launcher/CMakeLists.txt` installs `launcher` and `kyty_emulator`, but its Qt deployment step is Windows-only and there is no `macdeployqt` equivalent for Apple. The resulting 6.6 MB artifact omits the Qt Concurrent/Network/Widgets frameworks and the Cocoa platform plugin, so the launcher cannot start on a clean Mac that lacks the runner's Qt installation. Deploy the Qt frameworks/plugins before upload and verify the installed launcher outside `$QT_ROOT_DIR` (for example with `otool -L` plus a smoke launch).

Nmzik (commented):
Re-review of ec449fd: the launcher compilation failure is resolved, and the macOS and Linux checks pass; Windows is still running. The earlier `/proc`-based stack-discovery finding remains unresolved. One additional blocking artifact-packaging issue is included inline.

PR #117, opened 2026-07-28
Linux: port the emulator to a working state on Linux

StefanosCosta:
## Summary

  Brings the emulator up on Linux. Previously the Linux paths were a mix of stubs,
  `ENOSYS` returns and fatal "unsupported on this platform" aborts; games did not run.
  This implements the missing platform layers and fixes the Linux-specific bugs behind them.

  Windows behaviour is unchanged throughout — every change is either in a Linux-only file,
  inside a platform guard, or build config gated on `LINUX`/`KYTY_CLANG_CL`.

  ## What's included

  **Memory and page tracking**
  - Implement page-fault invalidation, protection query and `mprotect` transitions in
    `pageManager` — these were `Fatal("unsupported on this platform")`, so the CPU/GPU
    coherency model did not function at all
  - Implement region-tracking thread identity via `gettid`
  - Constrain the guest heap to the window `sceLibcMspaceCreate` accepts, so `mspace`
    creation stops returning null and faulting on the first `operator new`
  - Split reservation records instead of overwriting them; release physical pages on
    decommit via `madvise`; reserve and commit in one call rather than ~131k syscalls per 2 GiB
  - Stop reporting a preserved placeholder where none exists, which turned a recoverable
    unmap failure into a fatal rollback

  **Host exceptions and signals**
  - Implement the SIGSEGV/SIGBUS/SIGILL handler that feeds guest fault resolution
  - Make the fault handler async-signal-safe (`QueryHostProtection` allocated inside it)
  - Deliver `sceKernelRaiseException` on its target thread; populate all 16 GPRs, rflags,
    cs and ss in the guest signal ucontext, with write-back so a guest handler's mutations apply
  - Emulate SSE4a and MONITORX/MWAITX, which previously died with SIGILL on Intel hosts

  **Guest threading**
  - Save r14/r15 across guest thread entry — guest entry points do not reliably honour the
    SysV callee-saved contract, and a dirty register on return corrupted the emulator

  **Sockets and networking**
  - Implement the BSD socket surface against POSIX sockets (previously all `ENOSYS`),
    plus `Select`, `EpollWait` via `poll`, and name resolution via `getaddrinfo`
  - Translate host errno to the guest's FreeBSD numbering, `SOL_SOCKET` option names, and
    `MSG_*` flags — without which guest `EAGAIN` arrived as `EDEADLK`, `SO_REUSEADDR` landed
    on `SO_ERROR`, and `MSG_WAITALL` read as `MSG_DONTWAIT`
  - Fix `NetInetNtop` passing the guest address family straight to `inet_ntop`

  **libhttp2** *(cross-platform)*
  - Stop reporting a fabricated "204 No Content" success for requests that are never sent.
    Callers took that as a real response and dereferenced the empty result. Now mirrors
    libhttp v1 in `network.cpp`: the send fails and every accessor returns the recorded result.

  **Files, time and diagnostics**
  - Fix `SysFileSize` truncating to 32 bits; resolve guest paths case-insensitively so assets
    mastered on a case-insensitive filesystem load on ext4; carry nanoseconds in timestamps
  - Implement stack walking, faulting-module reporting, and the RenderDoc capture path
  - Use `__rdtsc` with TSC calibration instead of a hardcoded 1 GHz, and add a high-resolution
    sleep path (a 2 µs request took 55 µs)

  **Build, tests and CI**
  - Enable exceptions for the three test targets that use `throw`/`try`; they could not compile
    off Windows despite having unconditional `add_test` entries
  - Register 5 previously Windows-only GPU tests, taking ctest from 10 to 15; the remaining 3
    stay gated with the measured reason
  - Link the system libraries the static ffmpeg archives need; deploy Qt next to the installed
    launcher; add the Linux CI workflow

  **Launcher**
  - Probe for a terminal emulator instead of hardcoding `gnome-terminal`, and report a failed
    launch instead of failing silently

  ## Platform impact

  - **Windows** — unchanged. Audited hunk by hunk: the `pthread` asm emits an identical
    instruction stream, `network.cpp` additions are all in `#else` branches, `pageManager`'s
    Windows paths are untouched, and the refactored IWYU flag string expands identically.
  - **macOS** — the Linux implementations of the host exception handler and thread identity
    are confined to Linux, since Darwin has neither glibc's `ucontext` layout nor `gettid`.
    macOS keeps the behaviour it had before this branch. It is not made worse, and no
    untested Darwin code is introduced.

  ## Testing

  Built and run on Linux (Clang, Ninja). Verified against Subnautica, Cult of the Lamb and
  Dreaming Sarah: they boot, render, and sustain tens of thousands of draw calls without faults.

  Windows and macOS were not run — see the platform notes above for how each is kept unchanged.


Nmzik:
@StefanosCosta Thanks for the PR. I trimmed some comments, added platform guards to keep the macOS and Windows paths unchanged, and refactored CI into a single workflow. Could youconfirm that the latest revision still runs all three tested games successfully on Linux? If so, I’ll be happy to merge it.

StefanosCosta:
Hi there @Nmzik! Thanks for the quick review! All seems to run well except subnautica because of the libNet changes you reverted, I understand that's probably because they weren't linux only and risked a blast radius.I can add them with guards to make sure windows functionality is not affected, but If you think they are out of scope, then feel free to merge right now and I will creatte another PR for it :). On the other hand, I could push another commit that you give your final say to and revert if not needed. Up to you either way.

Nmzik:
Merged. Yeah lets split this PR. The ABI changes should go in a separate PR. Feel free to create one. Thanks for understanding

StefanosCosta:
Of course. It's better practice anyway. Opening it soon. Thanks!

PR #122, opened 2026-07-29
Drop redundant PROT_NONE tracking in reserve paths for Linux

claxten10:
* Some UE4 games, such as The Pathless, reserve a 512 GiB virtual address range during libc startup. Tracking every 4 KiB page causes a long delay and is unnecessary since the range is already PROT_NONE, and untracked pages are treated as NoAccess.


Nmzik (approved):

Nmzik:
Thanks for PR

PR #124, opened 2026-07-29
Optimize bulk memory invalidation

Nmzik:
- Replace per-page host-write fault handling with range cache invalidation.
- Preserve GPU-owned buffer and image data before file reads overwrite guest memory.
- Track mapped ranges for fast full-range checks.
- Add large-range ownership regression coverage.

Tests:
- `page_manager_tests`
- `memory_tracker_tests`
- `shader_recompiler_compute_tests`


Nmzik (commented):
Blocking finding: bulk invalidation now skips mapped pages whenever the destination range is only partially GPU-mapped.

PR #126, opened 2026-07-29
Emulate SHA-NI and fix SSE4a EXTRQ/INSERTQ register form

neobugs1:
I was testing games on KytyPS5 on my Intel i7-10700. But since Intel 10th gen CPUs and prior don't support SHA-NI, a game I was trying would crash with an illegal instruction like SHA1MSG1 before it could get anywhere.

This adds software emulation for the SHA-NI instructions in x64InstructionEmulator.cpp, using the same illegal-instruction path Kyty already uses for EXTRQ and MONITORX. If the host CPU has SHA-NI natively, none of this runs, so it shouldn't regress anything.

While testing I also fixed the existing SSE4a EXTRQ/INSERTQ handler. It only handled the immediate encoding and always advanced RIP by 6 bytes. The register form is 4 bytes and gets length/index from the source XMM register, so the old code could desync the instruction stream on Intel.

It made a difference on:
- PPSA10264
- PPSA20955

### DIFF
--- a/src/loader/x64InstructionEmulator.cpp
+++ b/src/loader/x64InstructionEmulator.cpp
@@ -522,44 +522,22 @@ static bool TryEmulateSse4a(PCONTEXT context) {
 		return false;
 	}
 
-	const uint8_t reg = ((modrm >> 3u) & 0x07u) | ((rex & 0x04u) << 1u);
-	const uint8_t rm  = (modrm & 0x07u) | ((rex & 0x01u) << 3u);
+	const uint8_t reg    = ((modrm >> 3u) & 0x07u) | ((rex & 0x04u) << 1u);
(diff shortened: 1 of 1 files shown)


Nmzik:
@neobugs1 Please include the game name or PPSA code (please edit PR comment) for SHA-NI, and I’ll happily accept the PR. Thanks for understanding.

Nmzik:
@neobugs1 Thanks for PR. There’s a definite regression in the INSERTQ changes on my Intel host. Let’s accept only the SHA-NI changes for now - I verified them on Windows and Linux(WSL2), and the results match the hardware 1:1. Could you please retest SHA-NI changes on your Intel host?

neobugs1:
Yeah I retested, this works too.

PR #129, opened 2026-07-29
Fix fabricated HTTP2 success

Nmzik:
Fix reporting a fabricated "204 No Content" success for requests that are never sent.

Thanks @StefanosCosta for finding the bug. Reproduced on my end.


StefanosCosta:
Glad to be of help!

PR #134, opened 2026-07-30
shader: specialize cube image descriptors

nikosszzz:
This PR adds cube image awareness to shader resource specialization and SPIR-V image sampling.

Image descriptors are now specialized with cube map information, validated against the runtime descriptor, and propagated into the IR. During SPIR-V emission, cube images that are represented as 2D array images have their sampling coordinates adjusted to match the expected coordinate space.

## Changes

- Added cube map detection from image descriptors (`DescriptorIsCube`).
- Extended resource specialization to:
  - Store whether an image resource is a cube map.
  - Validate cube state during specialization.
  - Propagate cube information into patched IR instructions.
- Added `cube`/`image_cube` fields to `ImageResource` and `MemoryInfo`.
- Updated SPIR-V image coordinate emission to apply the required coordinate offset for cube images emitted as `Dim2DArray`.

## Why

Previously, specialization only tracked the image dimension. Cube maps were treated the same as ordinary 2D array images, preventing the SPIR-V backend from distinguishing them when generating image sampling instructions.

By preserving cube information throughout specialization, the SPIR-V emitter can correctly handle cube image sampling while leaving the behavior of non-cube image resources unchanged.

## Testing

- The Minecraft Panorama on the Main Menu was suffering from this issue, now it works.
- Tested Dreaming Sarah, Subnautica and The Pathless to make sure my changes didn't break another game.

### DIFF
--- a/src/graphics/shader/recompiler/emitter/spirvEmitterImageHelpers.cpp
+++ b/src/graphics/shader/recompiler/emitter/spirvEmitterImageHelpers.cpp
@@ -2,6 +2,66 @@
 #include "graphics/shader/recompiler/emitter/spirvEmitterInternal.h"
 
 namespace Libs::Graphics::ShaderRecompiler::Spirv::Emitter {
+namespace {
+
+uint32_t EmitCubeAxisF32(EmitterState& state, uint32_t value) {
+	const auto normalized = state.builder.AllocateId();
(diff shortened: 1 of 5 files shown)


Nmzik:
@nikosszzz Thanks for finding the root cause. Accepted!

PR #136, opened 2026-07-31
macOS: enable guest signal dispatch on the target thread

m1abdullahh:
## What this fixes

Unity (IL2CPP) titles freeze on macOS at their first garbage collection. The GC does a stop-the-world: it raises a signal at every managed thread and each thread's handler parks its own thread until the collector resumes it. On macOS, `KernelRaiseException` ran the guest handler on the calling thread instead of the target, so the collector parked itself and the game froze. Metal Slug Tactics (PPSA20643) showed this as a black screen frozen at about frame 167.

The Linux port (#117) already added correct pthread_kill based delivery, but it is compiled out on macOS. This PR enables the same path there:

- translate between the Darwin mcontext (`uc_mcontext->__ss`) and the guest ucontext in `CreateSignalUcontextFromHost` / `ApplySignalUcontextToHost`
- use SIGUSR1 as the host dispatch signal, since macOS has no realtime signals
- block the dispatch signal inside the host fault handler, so a suspend request cannot preempt fault resolution between the protection fix and the retry

## Tested

On an M3 Pro MacBook Pro:

- Metal Slug Tactics: previously froze at the first collection on every run. Now runs through GC into menus and gameplay loading on every run, around 22 fps. It still aborts later when a guest AVX store hits a write tracked page (Rosetta's EmulateForward cannot handle AVX). That is a separate pre-existing issue, related to the TryEmulate note from #102, and I am looking into it independently.
- Raiden III: unaffected, runs at 60 fps on current main.

Windows and Linux code paths are unchanged.

### DIFF
--- a/src/common/hostException.cpp
+++ b/src/common/hostException.cpp
@@ -324,6 +324,10 @@ bool InstallHandler(Handler handler) {
 	sa.sa_sigaction = SignalHandler;
 	sa.sa_flags     = SA_SIGINFO;
 	sigemptyset(&sa.sa_mask);
+	// The guest signal-dispatch path (KernelRaiseException) interrupts threads with
+	// SIGUSR1; block it while a fault is being resolved so a stop-the-world request
+	// cannot preempt the handler between the protection fix and the retry.
+	sigaddset(&sa.sa_mask, SIGUSR1);
 
 	// macOS raises SIGBUS for protection faults on some paths and SIGSEGV on others;
 	// SIGILL covers instructions the host cannot execute (routed to the x64 emulator).
--- a/src/libs/libKernel.cpp
+++ b/src/libs/libKernel.cpp
@@ -40,7 +40,10 @@
 #define NOMINMAX
 #endif
 #include <windows.h>
-#elif !defined(__APPLE__)
+#elif defined(__APPLE__)
+#include <csignal>
+#include <sys/ucontext.h>
+#else
 #include <csignal>
 #include <ucontext.h>
 #endif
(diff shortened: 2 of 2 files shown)


Nmzik:
@m1abdullahh WOW. Thanks for the quick response! Are you sure you tested it with the latest code? :) I’m amazed it still works on macOS! :D https://github.com/KytyPS5/KytyPS5/commit/2f5396c6a536374869c9edc4ff32102464616930

Nmzik:
@m1abdullahh This is APPLE guarded code. Thanks!

m1abdullahh:
@Nmzik  oh, you merged it already! nice! This PR was actually for another issue I encountered while running Metal Slug Tactics and this is a fix for that and this has been WIP for some days.

But, yeah, just before creating this PR and after we spoke on email, I did test this on top of latest code with both Raiden III and Metal Slug Tactics and I see MASSIVE fps improvements! great work there!

m1abdullahh:
Actually, correction to what I just wrote: that testing was from just before #135 landed, so not quite the latest code. Prompted by your question I re-ran everything on current main including #135, and here is the accurate picture:

- This PR's fix works the same on latest main. Metal Slug Tactics gets through GC into menus and loading on every run.
- The fps improvements are real, Raiden III went from about 2.5 fps right after the renderer rewrite to a locked 60 fps.
- However #135 seems to have introduced a macOS regression: Raiden III now crashes about 5 seconds after boot with a guest null deref (Read at offset 0x38, right after Audio init). It looks like a timing sensitive startup race: 6 out of 6 crashes with --printf-direction Silent, but it never reproduces with Console logging or even a slightly perturbed binary, so it hides from debugging. Bisected: ecb48f9 is fine (4/4 runs), 2f5396c crashes (6/6 runs). Metal Slug Tactics is not affected.

Happy to dig further on real hardware, just tell me what you would like logged.

Nmzik:
@m1abdullahh Thanks for testing the latest major change in #135. At least some games are running on the latest code.

m1abdullahh:
I'll dig into the Raiden regression on my hardware and report back with a root cause. Since it hides from any instrumented build I'll have to get creative, so it may take a bit.

PR #140, opened 2026-07-31
launcher: add per-test log output option

Digote:
## Description

While testing games, I often needed to preserve the log from each individual run. The existing `File` option always uses the same output path, so I had to manually rename the log before starting another game or test to prevent it from being overwritten.

This PR adds a new `File per test` option to the launcher's `Printf direction` setting. When selected, the launcher preserves the configured directory, base filename, and extension, but appends the game ID and a timestamp to the filename.

For example:

`_kyty_CUSA12345_20260731_153012_421.txt`

The existing `Silent`, `Console`, and `File` options remain unchanged. This is intentionally a small and simple improvement, since the logging system may receive more extensive changes in the future.

## Image
log

log2

## Testing

- Built and tested locally on Windows.
- Verified that `File per test` creates a new log for each run.
- Verified that the regular `File` option keeps its existing behavior.
- Verified that the setting is saved and restored by the launcher.

## AI assistance

AI assistance was used to analyze the existing logging flow, implement the initial changes, and review the final diff. I reviewed the code, compiled it, tested the behavior locally, and take responsibility for the submitted changes.

### DIFF
--- a/src/launcher/forms/configuration_edit_dialog.ui
+++ b/src/launcher/forms/configuration_edit_dialog.ui
@@ -203,7 +203,7 @@
       <item row="7" column="1">
        <widget class="QComboBox" name="comboBox_printf_direction">
         <property name="toolTip">
-         <string>Print logs to file or console window. If enabled may decrease emulator performance</string>
+         <string>Print logs to the console or a file. File per test creates a new file for every test</string>
         </property>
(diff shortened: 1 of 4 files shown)


Nmzik:
Unnecessary. You can always redirect output

PR #145, opened 2026-08-01
fix: fix Linux memory issues, renderer synchronization, shader decoding, and TLS cleanup

nikosszzz:
## Summary

This PR includes several fixes across the Linux memory manager, renderer, shader decoder, and libc runtime.

### Changes
- Reserve only available guest address ranges on Linux.
- Add support for the `S_BFE_I32` shader instruction.
- Preserve GPU image contents during CPU writes.
- Correctly execute thread-local storage destructors on thread exit.

# Fixed issues
- Minecraft had a crash when saving & quitting a world. The thread-local changes have fixes this.
- Minecraft had a text corruption bug in the new so-called Ore UI, wherever it is used. The text would become corrupted after selecting the Create World button or navigating the menus for some seconds. Example of corruption: Create world -> just showed "a" after the corruption. Going in and out of the menu would fix it.
- Minecraft also again exposed another issue, a missing opcode. This time in the world creation menu, if you selected the Creative button, it would crash with an error for missing the opcode handling for `S_BFE_I32`. This has fixed it.
- One of the recent memory commits had affect Linux and effectively stopped the emulator working or quitting some games too early. The fix and issue reminds me of the macOS memory issue from some days ago.

# Games tested
Minecraft
Pathless
Dreaming Sarah

### DIFF
--- a/src/graphics/host_gpu/renderer/depthRenderTarget.cpp
+++ b/src/graphics/host_gpu/renderer/depthRenderTarget.cpp
@@ -96,9 +96,11 @@ void RenderExecutor::ResolveRenderDepthTarget(uint64_t submit_id, RenderCommandB
 	const auto& dc = hw.GetDepthControl();
 	const auto& sc = hw.GetStencilControl();
 	const auto& sm = hw.GetStencilMask();
-	const bool  depth_active =
+	const bool  has_stencil =
+	    z.stencil_info.format != Prospero::GpuEnumValue(Prospero::StencilFormat::kInvalid);
+	const bool depth_active =
 	    dc.z_enable || dc.z_write_enable || dc.depth_bounds_enable || rc.depth_clear_enable;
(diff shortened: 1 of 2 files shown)


Nmzik:
@nikosszzz Thanks for the PR! The library and memory changes have been accepted. Ill find a universal solution for the renderer fix, and BFE opcode doesn’t work correctly on NVIDIA GPUs. If this test works for you, it means you have an AMD GPU.
No further work is needed from you. I’ll close this once I find the right solution. Thanks for finding these bugs!

nikosszzz:
> @nikosszzz Thanks for the PR! The library and memory changes have been accepted. Ill find a universal solution for the renderer fix, and BFE opcode doesn’t work correctly on NVIDIA GPUs. If this test works for you, it means you have an AMD GPU.
> 
> No further work is needed from you. I’ll close this once I find the right solution. Thanks for finding these bugs!
> 
> 

Yep, I'm on RDNA2 pretty close to the PS5.

PR #146, opened 2026-08-01
Fixes for Linux and UE4 games

claxten10:
* This PR fixes a deadlock spotted on Linux
* Fixes crashes that can impact other titles as well.

* With this PR, an UE4 title "The Pathless" successfully reaches the Main Menu

* Only tested on Linux. Windows and macOS are untested, though no regressions should appear on either.

* The Pathless, Dreaming Sarah, Subnautica and Minecraft have been tested, no regressions have been spotted with this PR.

image


Nmzik:
@claxten10 Thanks for the PR. No further work is needed from you, thanks for splitting the changes into separate commits.

1. I’ll take a look at this.
2. Accepted.
3. Accepted.
4. This doesn’t map 1:1 to Vulkan. Need to check
5. Accepted.
6. This is a bug in the current implementation. I’ll fix it.
7. Fixed 84236d1
8. This is not correct.

PR #156, opened 2026-08-03
Unity + linux: signal safety, direct-memory zeroing, AudioOut2 pacing, renderer and SaveData fixes

StefanosCosta:
# Summary

Fixes deadlocks, stalls and correctness bugs found while getting FMOD-based titles to a playable
state. The two largest are in guest signal delivery and audio submission: a guest signal handler
could run while the emulator held its own locks and deadlock the process, and AudioOut2 was paced
against a modelled queue rather than the device, so the device ran dry 98% of the time.

These are platform-neutral changes in shared code, not Linux-only — Windows is affected and was not
run. See Platform impact for more details.

This PR has major fixes for Unity games. 

## What's included

Guest signals and condition variables

    CondVar::SignalThread woke its targets while holding the waiter-registry mutex; waiters retire
    through UnregisterCondWaiter, which needs that same mutex, so the waker deadlocked against
    everything it was trying to wake. Collect, release, then wake.
    Fixing that opened a use-after-free — a waiter can unregister and destroy its CondVar between
    the collect and the wake. The registry now holds shared_ptr. Confirmed under ASan with the
    defect reintroduced; tests added.
    Defer guest signal delivery on non-guest frames and inside emulator critical sections.
    sceKernelRaiseException only carries POSIX_SIGUSR1 in practice — IL2CPP's stop-the-world GC
    handler, which blocks until collection ends; running it on a frame holding an emulator lock
    hangs the process permanently.
    Document MemoryOperationLock's member order, which is load-bearing: members destroy in reverse,
    so the mutex must release before the critical section is left.

Audio

    Pace AudioOut2ContextPush by the device instead of a wall-clock queue model — it fed at exactly
    1x realtime with no cushion. Was 98% device underruns with the queue capped at two buffers; now
    0.01% at the full 40 ms target.
    Implement the sceKernelSyncOnAddress wait/wake pair. Both NIDs currently bind to one stub that
    never compares the word, blocks or wakes, so FMOD's handshake never completes — Cat Quest III
    hangs at its first flip (EndOfPipe count 1 over 45 s, reproduced twice; 4849 once implemented).
    Adopt upstream's disassembly-derived AudioOut2ContextParam layout over the one this branch
    carried. The guest's raw block reads 18/128/0/1/256/1, which only upstream's fields explain —
    and reading it correctly is what exposed the underrun above, since the guest asks for
    queue_depth = 1.

Memory

    Zero freshly allocated direct memory — allocations returned the previous owner's bytes. FMOD
    reads a buffer before writing it, and stale NaN bytes propagate through FMA chains without
    raising an FP exception, poisoning the DSP state for the rest of the process. Zeroed at
    allocation, not at map time, so map/unmap/remap still preserves contents. Same gap in the
    pool-expansion path.
    Evidence: same binary one env var apart, 248 s clean versus NaN at 61.3 s, on two titles with
    different audible symptoms.
    This re-proposes work not taken from #147. Without the signal and audio fixes here, titles do
    not reach the in-game state where the stale allocation is read, so it may not have reproduced in
    isolation.

Renderer

    Fix four sources of invalid draws: null-image keying in the texture cache, depth-compare state
    applied to a colour surface, colour-write-enable left set for depth-only passes, and a render
    area wider than its attachment.

RenderDoc

    Pass the Vulkan dispatch key rather than the VkInstance handle — RenderDoc defines it as the
    first pointer inside the dispatchable object, so passing the handle makes StartFrameCapture
    silently do nothing.
    Adopt an already-open capture instead of resetting to Idle; nothing then called
    EndFrameCapture, so the capture stayed open and no file was written.
    Linux only, in an #else branch: X11 sessions were misreported as Wayland and capture refused
    outright, because the source-tree SDL_config.h shadows the generated one.

SaveData, filesystem and loader

    Persist sceSaveDataSetParam's arguments, which were discarded — saves enumerate as damaged
    Mount /devlog used by Unity games; fix a lock leak on the lseek path
    Remove a debug dump in the access-violation handler keyed to a hardcoded guest address

Diagnostics

    Add an opt-in HLE call trace behind KYTY_ENABLE_HLE_TRACE, off by default and a no-op when
    disabled

## Platform impact

    Not platform-guarded. Every change except the RenderDoc X11 branch is in shared code and
    affects Windows.
    Windows was not run — no Windows machine here. The CondVar/HleCriticalSection and signal
    deferral work is what most deserves a Windows eye, since it interacts with host thread
    suspension.
    macOS rides the Linux/POSIX paths; nothing here is Darwin-specific.

## Testing

Built and run on Linux (Clang, Ninja). Every commit builds standalone; ctest 6/6;
-DKYTY_ENABLE_HLE_TRACE=ON builds.

Boot health as EndOfPipe Signal count over a 45 s run (1 = hung): PowerWash Simulator 3906,
Cat Quest III 4740, Subnautica 2764.

Audio verified by ear — Subnautica, PowerWash Simulator, Cult of the lamb  menu and in-game,  Superliminal menu. No
clicking or dropouts, and audio survives sustained in-game movement, which previously killed it
permanently. Not machine-checked: the glitch counter gave 31.9/s and 87.6/s for the same binary.

AI was used for investigation, bisection and drafting; every claim above is backed by a measurement
or by my own listening test.

The following titles were either not playable  before these changes:

1. Subnautica
2. Powerwash Simulator
3. Cult Of the Lamb
4. Cat Quest 3

Subliminal did not boot to the menu as well.

Fixes have been confirmed to be universal and not game-specific, and as such were pushed for review.

Proof of Errors and Working In-Game footage can be found here:

https://drive.google.com/drive/folders/1tpYUxadOKHL7QwwBMez_lWwphPQ7adYB?usp=sharing

 Note on commit bb75992: its subject mentions implementing the address-wait pair, but that half is no longer in the branch. 
 Upstream's 00a5dd3 landed a fuller sceKernelSyncOnAddressWait/Wake (32/64-bit variants, dedicated src/kernel/syncOnAddress.cpp,  its own tests), so the merge of upstream/main resolved in favour of upstream's implementation. All that remains from that commit is the AudioOut2 device-paced submission.


StefanosCosta:
Hey @Nmzik . Did my best to make it easy for you to cherry-pick whatever you want. Both clicking and audio cut-off issues from previous PR are fixed in this PR.

StefanosCosta:
Hey @Nmzik. I realize that many of the changes in this branch are obsolete given the latest changes merged in main, and probably not welcome as a relatively large PR. It is not my intent to make things difficult or inconvenient. I separated the remaining valid commits into 6 other branches which should make it easier for you to review, and excluded the changes that are now obsolete. However, I don't want to cut you off if you already started cherry-picking any commits, so I will close this PR and open the rest of them only after receiving your OK. Thanks again for your amazing work

Nmzik:
@StefanosCosta Just so you know, I’m actively reviewing your PRs - not ignoring them. However, please dont expect me to merge them immediately, as Im also working on other things.
I also dont want to merge hacks into the main codebase. If you’ve discovered an actual bug, thats fine, and I’ll credit you for finding it. However, some of the changes look more like workarounds/hacks than proper implementations that will work across all games. Thanks for understanding. Not trying to be mean

StefanosCosta:
@Nmzik Of course. You're right. That's what I realized, that's why I want to change it. I'm not pushing or complaining about it. I'm mostly ashamed because I didn't actually give you something worth working with and honestly if I was in your position I would have rejected it. I just wanted to make sure I didn't interrupt you if you actually did start looking through it.

Nmzik:
@StefanosCosta It’s all good - don’t close this PR, as some parts of it still need further investigation. However, if you find other unrelated bugs in the future, please submit them as separate, smaller PRs, even if each fix only helps games progress a little further. You can reference related PRs in each other so that I can understand the changes more easily.

StefanosCosta:
@Nmzik Thank you. I will stick to that practice from now on

gandalfthewhite19890404:
Why not mergin this?
Or cherry pick some commits from PR - save fixes, i.e.
For what big PRs, why not separate to small - it easier for testing and merge can be completed faster

StefanosCosta:
@gandalfthewhite19890404 As you can read from the previous comments some of these fixes are hacks. The file save fixes and others have already been committed in separate commits by the maintainer. I was going to split it into smaller PRs but since they already started investigating we concluded to leave this as is and make smaller PRs from now on. At least that's how I understood it

Nmzik:
`Persist sceSaveDataSetParam's arguments, which were discarded — saves enumerate as damaged `

I understand that this is a real bug. Is there any game that actually checks or relies on this behavior? I don’t know of any game that really depends on it.

Nmzik:
`Fix four sources of invalid draws: null-image keying in the texture cache, depth-compare state
applied to a colour surface, colour-write-enable left set for depth-only passes, and a render
area wider than its attachment.`

Which game caused this change?

StefanosCosta:
Hi @Nmzik ! Both issues that you just mentioned were identified in subnautica. With main changes a couple days ago, I was able to replicate only 2 of the 4 sources of invalid draws. Both showed up I believe in both subnautica and powerwash simulator. I believe the save issue also exists both in subnautica and cat quest 3.

Nmzik:
@StefanosCosta Could you please confirm whether the following games are working on Linux with the latest main branch?

* Subnautica
* PowerWash Simulator
* Cult of the Lamb
* Cat Quest III
* Subliminal

Nmzik:
> Hi @Nmzik ! Both issues that you just mentioned were identified in subnautica. With main changes a couple days ago, I was able to replicate only 2 of the 4 sources of invalid draws. Both showed up I believe in both subnautica and powerwash simulator. I believe the save issue also exists both in subnautica and cat quest 3.

Definitely not "cat quest 3". It does save

StefanosCosta:
The issue was i believe corrupted save files, so it wouldn't boot after saving. Checking now

Nmzik:
> The issue was i believe corrupted save files, so it wouldn't boot after saving. Checking now
> 
> 

Yeap, that was the issue but it was fixed

StefanosCosta:
Hey @Nmzik . In current main, none of the titles go in-game. Subnautica goes in menu but never goes in-game. I believe the changes in this branch solved that issue, though I don't remember which ones. The same was true for powerwash simulator, and cult_of_the_lamb (they don't reach the menu, but superliminal does). I went into my branch that this PR is from and they run fine there. Would you like me to isolate the changes that make them run?

StefanosCosta:
@Nmzik  Forgot to mention cat quest 3 works fine. My bad. Rest of them still have an issue  though

StefanosCosta:
@Nmzik  Sorry for the constant pinging. I just checked my notes because I kept a backlog. Commits 12f71e0 + 1d25e35 made subnautica reach in-game instead of hanging forever. After that If i remember correctly the rest of the games also reached not only the menu but also in-game.  I'm starting a test branch and cherry-picking them in there to test and confirm for the other titles, but consider it true for Subnautica. Regarding the save files, subnautica also had corrupted game files, so I'm not sure if it's affected by what you mentioned previously

StefanosCosta:
Commits that made the games Cult of the lamb, powerwash simulator and Subnautica run and reach in-game  were separated and submitted in #215.

Nmzik:
Cult of the lamb, powerwash simulator and Subnautica seems to be working now on Linux Feel free to open the issue

PR #161, opened 2026-08-04
Report Vulkan validation errors instead of ending the process

Death-Whispers-U:
`--vulkan-validation` decides whether validation runs. It also decides whether a spec violation kills the emulator, and those are two different questions.

The debug callback calls `EXIT_COLOR` on any validation-type error, so the first violation ends the session whatever the title was doing. The launcher turns validation on by default (`vulkan_validation_enabled = true` in `configuration.h`), so this is the configuration most people are actually in — the option is effectively unusable for anyone who just wants the diagnostics.

Failing fast is useful when you are working on the emulator itself. It is not a good default for someone trying to run a game.

## Change

Drop the abort. Errors are never skipped by the filter just below, so every message still reaches the log — only the process death is gone.

## Testing

PAC-MAN WORLD Re-PAC with validation on:

| | before | after |
| --- | --- | --- |
| result | dies | runs the whole window |
| log lines | 31,460 | 31,668,560 |

It still reports the violations it hits (10x VUID-vkCmdDrawIndexed-None-07749, 10x VUID-vkCmdBeginRendering-pRenderingInfo-09588 in that run) — it just does not die on them. I have separate changes for both of those violations.

### DIFF
--- a/src/graphics/presentation/window/vulkanWindow.cpp
+++ b/src/graphics/presentation/window/vulkanWindow.cpp
@@ -754,10 +754,14 @@ static VKAPI_ATTR vk::Bool32 VKAPI_CALL VulkanDebugMessengerCallback(
 		default: severity_str = "?";
 	}
 
-	if (error) {
-		EXIT_COLOR(severity_style, "[Vulkan][%s][%u]: %s\n", severity_str,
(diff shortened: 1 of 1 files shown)


Nmzik:
@Death-Whispers-U Let’s just disable Vulkan validation for now

PR #163, opened 2026-08-04
Keep a sampled depth attachment in a layout both uses agree on

Death-Whispers-U:
When a draw uses the same depth image as its depth attachment and as a sampled texture, the layout recorded for `vkCmdBeginRendering` is already stale by the time the draw happens:

```
vkCmdBeginRendering(): pDepthAttachment is expected to have layout
DEPTH_STENCIL_ATTACHMENT_OPTIMAL but previous known layout is
DEPTH_STENCIL_READ_ONLY_OPTIMAL
(VUID-vkCmdBeginRendering-pRenderingInfo-09588)
```

The order in the draw path is:

```
AcquireRenderTargets   transitions the depth image and records that layout
CommitBindings         transitions it again, to a read-only layout, because it is sampled
BeginRendering         declares the layout recorded in the first step
```

## Fix

The colour path already solves exactly this: the attachment uses `eGeneral` when the image is also bound as a shader resource, and the descriptor path uses `eGeneral` for images that are render targets. Depth had neither check, and the descriptor branch excluded depth outright.

This mirrors the colour behaviour for depth. Access flags are now picked by aspect, since that shared branch previously assumed colour.

I used `force_general` rather than `is_target` deliberately: `is_target` also drives image merging in the texture cache, and this change has no business reaching there. `force_general` is per-draw state that `ResetBindings` clears.

## Testing

STORY OF SEASONS with validation on:

| | before | after |
| --- | --- | --- |
| result | dies on the VUID | survives the full window |
| log lines | 503,751 | 46,887,049 |
| validation errors | 2 | 0 |

PAC-MAN WORLD Re-PAC and Double Dragon Gaiden show no regression.

Re-verified after rebasing onto current main, since two recent renderer commits touch this path.

### DIFF
--- a/src/graphics/host_gpu/renderer/pipeline/descriptors.cpp
+++ b/src/graphics/host_gpu/renderer/pipeline/descriptors.cpp
@@ -1135,15 +1135,19 @@ void RenderExecutor::CommitBindings(CommandBuffer&                     buffer,
 				                            vk::AccessFlagBits2::eShaderWrite
 				                      : vk::AccessFlagBits2::eShaderRead,
 				              range, vk_buffer);
-			} else if ((image.binding.force_general || image.binding.is_target) &&
-			           !image.info.IsDepth()) {
+			} else if (image.binding.force_general || image.binding.is_target) {
+				// Depth reaches this branch too: AcquireRenderTargets sets force_general on a
+				// depth attachment the same draw samples, and both sides must agree on eGeneral.
+				const auto attachment_access =
+				    image.info.IsDepth() ? vk::AccessFlagBits2::eDepthStencilAttachmentRead |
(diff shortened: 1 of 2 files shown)


Death-Whispers-U:
Rebased onto `c52bf450`. The conflict was mechanical: `CommitBindings` now loops over `prepared_bindings` rather than handling a single `prepared`, so the fix moves into the loop body unchanged.

The bug is still present on today's main:

- `descriptors.cpp:1139` still excludes depth from the general-layout branch with `!image.info.IsDepth()`
- `renderDraw.cpp:667` still takes `depth_attachment_layout(depth)` unconditionally

Linux, Windows and macOS are green. +20/-6 across 2 files.

PR #167, opened 2026-08-04
Decode sRGB in the shader for formats the host image cannot carry

Death-Whispers-U:
PAC-MAN WORLD 2 Re-PAC renders its cutscenes magenta. Everything is there — geometry, detail, motion — but drowned in pink.

`R8_SRGB` and `R8G8_SRGB` are optional in Vulkan and plenty of devices expose no features for them at all, so `k8Srgb` and `k8_8Srgb` are created as unorm images. That part is already here and is the right call. What is missing is the other half: nothing then supplies the sRGB decode those formats are owed, so the sampled texels stay gamma-encoded and any title that binds one gets a colour cast.

PAC-MAN WORLD 2 binds exactly these — its decoded-video planes are `k8Srgb` luma and `k8_8Srgb` chroma.

## Change

Emit the decode in the shader.

`IsShaderDecodedSrgbTextureFormat` names the two formats. It is deliberately not a column in the format table, because this is a property of what the host can express, not of the guest format: every other guest sRGB format has a real host counterpart whose sampler already decodes it, and a shared column would sweep those in and decode them twice.

Specialization carries it. `ResourceMaterialization` fills `srgb_decode` in from the runtime descriptor and rejects a descriptor that no longer matches the permutation that was compiled, and the flag joins the program key — two modules differing only in the decode produce different SPIR-V and must not collide. The uint check cannot stand in for it: `k8UNorm` and `k8Srgb` are both non-uint and indistinguishable to it.

Three details worth flagging for review:

- The decode runs **before** the bitcast, since the bitcast reinterprets the float as an integer bit pattern for VGPR storage and decoding after it would corrupt the value.
- Alpha is never decoded, matching Vulkan's own sRGB sampling semantics.
- `Gather4` returns one channel of four neighbouring texels, so it decides the alpha exclusion once, on that channel.

## Testing

Same cutscene frame, same second, mean channel values:

| | R | G | B |
| --- | --- | --- | --- |
| before | 239.0 | 88.1 | 253.2 |
| after | 68.8 | 79.6 | 106.3 |

Red and blue pinned high with green far below them is magenta, numerically. Green minus the average of red and blue goes from −158.0 to −7.9.

Verified again after rebasing onto current main.

### DIFF
--- a/src/graphics/guest_gpu/gpu_format.cpp
+++ b/src/graphics/guest_gpu/gpu_format.cpp
@@ -216,6 +216,18 @@ bool IsUintTextureFormat(BufferFormat format) {
 	return info != nullptr && info->uint_texture;
 }
 
+bool IsShaderDecodedSrgbTextureFormat(BufferFormat format) {
+	// Deliberately an explicit two-case test rather than a column in the format table, because
+	// this is not a property of the guest format on its own - it is a property of what the host
+	// can express. k8Srgb and k8_8Srgb are created as unorm images (see the comment on their
+	// entries in vulkanCommon.cpp) because R8_SRGB and R8G8_SRGB are optional in Vulkan and
+	// plenty of devices expose no features for them at all, so their sRGB decode is owed by the
+	// shader. Every other guest sRGB format has a real host counterpart - k8_8_8_8Srgb maps to
+	// eR8G8B8A8Srgb and each BC*Srgb to a native sRGB block format - and the sampler already
+	// decodes those. A shared table column would sweep them in and decode them twice.
+	return format == BufferFormat::k8Srgb || format == BufferFormat::k8_8Srgb;
+}
+
 BufferFormat RemapTextureFormat(BufferFormat format) {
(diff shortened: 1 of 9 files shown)


Death-Whispers-U:
Rebased onto `c52bf450`. Windows and macOS are green, Linux is still running.

The diff is larger than when this was opened, +242 rather than +130, because main grew indirect image tables in the meantime. One SPIR-V module is compiled for a whole table, so two candidates that disagree about `srgb_decode` must not share one. The flag now threads through `SpecializeResources` and the null-indirect-root check as well. The added test carries two negative controls, so it fails if the flag ever stops discriminating.

One thing I would rather flag than leave implicit. For a plain sampled image with no conversion format, `shader_swizzle` stays identity and the guest `DstSel` is applied by the host image view. The alpha exclusion therefore operates on post-swizzle slots, while Vulkan decodes before component mapping. A descriptor that swizzled the single sRGB channel into slot 3 would go undecoded. I have not seen a title do it, and correcting it needs an inverse-swizzle decision I would rather make with review input than silently.

PR #193, opened 2026-08-07
Makes launcher build optional

boykopovar:
Adds a `KYTY_BUILD_LAUNCHER` CMake option (ON by default). When disabled, the launcher is excluded (only launcher) from the default build, allowing the rest of the project to be built without Qt6. Previously, the only way to do this was to explicitly specify every non-launcher target on the build command line.

The option is enabled by default, so existing developer configurations and CI remain unaffected.


Nmzik:
Use cmake -S . -B build -DKYTY_BUILD_LAUNCHER=OFF

PR #197, opened 2026-08-08
feat(media): play self-decoded movies on the host

Death-Whispers-U:
Some titles decode video inside their own executable rather than through sceAvPlayer or libSceVideodec. Demon's Souls does: its logo, attract movie and title screen are 4K Bink 2 (.bk2) files decoded on the game's own CPU threads. No HLE export sees those frames, so what reaches the presenter is whatever the guest decoder produced, and the intro renders as confetti.

Identify the active movie from the kernel file open and decode it on the host instead, presenting those frames in place of the guest's.

ffmpeg is bound at runtime from KYTY_BINK_FFMPEG_DIR rather than linked, because the bundled prebuilt is configured without the binkvideo2 decoder. Binding by name keeps this an optional capability that degrades to the old behaviour when the libraries are absent, and needs no dependency change.

Off unless KYTY_BINK_HOST is set. The launcher enables it and points at a "plugins" folder beside the emulator when one is present; anything already in the environment wins.

Note the presented image is A2B10G10R10 for this title, not 8-bit RGBA, so the decode target is chosen from the swapchain format.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -185,6 +185,8 @@ file(GLOB kyty_emulator_src CONFIGURE_DEPENDS
 	"${KYTY_SOURCE_DIR}/libs/*.h"
 	"${KYTY_SOURCE_DIR}/graphics/*.cpp"
 	"${KYTY_SOURCE_DIR}/graphics/*.h"
+	"${KYTY_SOURCE_DIR}/graphics/media/*.cpp"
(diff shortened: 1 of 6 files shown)


Nmzik:
@Death-Whispers-U sorry, this is a hack. I won’t accept it. Video will be properly fixed in future PRs

PR #215, opened 2026-08-08
Unity Game Fixes - common/kernel: deliver guest signals only at safe points

StefanosCosta:
## NOTE

These are commits from PR #156, compressed and cleaned up to be isolated from the rest of the changes that have either a) been committed or b) been identified as hacks/incomplete implementations.

These changes make Subnautica, Powerwash Simulator and cult of the lamb boot and enter in-game. I do not claim that these changes are objectively the correct implementation. This is just a PR separating commits from #156 so that they can be seen in the correct context and provide better insight as to what their purpose is. More details below:

## Summary

  A guest signal handler could run nested inside a host wait, or while the thread held an
  emulator-global lock. Either one deadlocks the emulator: the handler blocks, the frame never
  unwinds, and the state it holds is stranded. This leaves the signal queued in both cases and
  dispatches it once the frame has unwound.

  ## What's included

  ### Guest signal delivery — `libKernel.cpp`, `memory.cpp`, `graphicsRun.cpp`

  - Leave a signal queued when it interrupted host code, and dispatch it from
    `KernelDispatchPendingSignalForCurrentThread` once the frame unwinds. Running IL2CPP's GC-suspend
    handler nested inside a host wait strands glibc's condvar group reference, after which every later
    broadcast on that variable blocks forever — the hard freeze on Subnautica's loading screen.
  - Declare the 33 `g_memory_operation_mutex` sites and the GPU submission lock as critical sections,
    and refuse to dispatch inside one. A handler that blocks while holding either stalls every other
    guest thread; both mutexes were confirmed owned by a single parked thread via `__data.__owner`.
  - Gate the dispatcher on one atomic load. It now runs on every wait and sleep exit, where the
    previous path did a read-modify-write per signal number.

  ### Condition variables — `threads.cpp`, `threads.h`

  - Wake `SignalThread`'s targets with the registry lock released. A broadcast can block until the
    waiters it is retiring have left, and they leave through `UnregisterCondWaiter`, which needs that
    same mutex — so the waker deadlocked against every waiter it was trying to wake. `Signal()` and
    `SignalAll()` already woke without the lock.
  - Hold registered waiters by `shared_ptr`. Between the collect and the wake, a waiter can return,
    unregister and have its owning `CondVar` destroyed, so the wake touched freed memory.
  - Add a dispatch point to `CondVar::WaitFor`. It had none, so a deferred signal sat undelivered for
    the whole of an arbitrarily long timeout.

  ### Build and tests — `CMakeLists.txt`, `tests/SignalSafetyTests.cpp`

  - Add `signal_safety_tests`, covering both the deadlock and the use-after-free. The use-after-free
    oracle is ASan: built with `-fsanitize=address` it reports `heap-use-after-free` in `WakeCondVar`
    reached from `SignalThread`, on memory freed by `~CondVar`, and is clean with the fix. A
    regression in the deadlock hangs the test rather than failing it, which is the honest shape for a
    deadlock.

  ## Platform impact

  - **Windows — changed, and not built or run here.** Two pieces are excluded by existing guards: the
    `CondVar::WaitFor` dispatch point sits in the non-`KYTY_WIN_CS` arm, and the
    `HostSignalDispatchHandler` deferral is inside
    `#if KYTY_PLATFORM != KYTY_PLATFORM_WINDOWS && defined(__x86_64__)`. Windows keeps its own APC
    path (`SignalApcHandler` / `nt_queue_apc_thread_ex`), untouched. Everything else compiles there:
    the `SignalThread` fix, the `shared_ptr` waiter lifetime, `HleCriticalSection`, the 33
    `MemoryOperationLock` sites, and the dispatcher's two new early returns.
  - **macOS** — rides the POSIX arm; compiles, not run.

  ## Testing

  Built on Linux (Clang, Ninja). `ctest` 27/27 including the new `signal_safety`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -433,6 +433,13 @@ add_executable(resource_tracking_tests EXCLUDE_FROM_ALL
 target_link_libraries(resource_tracking_tests fmt::fmt)
 target_include_directories(resource_tracking_tests PRIVATE ${inc_headers})
 
+add_executable(signal_safety_tests EXCLUDE_FROM_ALL
+	"${KYTY_TESTS_DIR}/SignalSafetyTests.cpp"
+)
+target_link_libraries(signal_safety_tests common)
+target_include_directories(signal_safety_tests PRIVATE ${inc_headers})
+
+
 add_executable(audio_out2_port_tests EXCLUDE_FROM_ALL
 	"${KYTY_TESTS_DIR}/AudioOut2PortTests.cpp"
 	"${KYTY_SOURCE_DIR}/libs/libAudio2.cpp"
@@ -508,6 +515,7 @@ if(BUILD_TESTING)
 	add_test(NAME shader_vertex_metadata COMMAND $<TARGET_FILE:shader_vertex_metadata_tests>)
 	add_test(NAME shader_stage_runtime COMMAND $<TARGET_FILE:shader_stage_runtime_tests>)
 	add_test(NAME resource_tracking COMMAND $<TARGET_FILE:resource_tracking_tests>)
+	add_test(NAME signal_safety COMMAND $<TARGET_FILE:signal_safety_tests>)
 	add_test(NAME event_queue_lifetime COMMAND $<TARGET_FILE:event_queue_lifetime_tests>)
 	add_test(NAME sync_on_address COMMAND $<TARGET_FILE:sync_on_address_tests>)
 	add_test(NAME audio_out2_port COMMAND $<TARGET_FILE:audio_out2_port_tests>)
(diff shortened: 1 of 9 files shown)


svoboda18 (approved):

Nmzik:
@StefanosCosta Is it still needed? All the games are running successfully on Linux, and the main development is now happening on Linux.  Powerwash Simulator is definitely goes in-game

StefanosCosta:
Hey @Nmzik. I checked your question around 3 days ago with the main of that day. I couldn't load the games on ubuntu far enough into the game to test if this is still needed. Subnautica didn't even boot to menu, though I guess it could be a problem on my side because powerwash simulator didn't boot for me either. I made sure to pull all recent changes from main though and created a new build. I'm going to be taking care of a few personal things so i'm going to be away from github for a while. Hope all goes well.

Nmzik:
Should be good on Linux

PR #230, opened 2026-08-10
Windows11 fix

Pouare514:
### Summary

Includes several fixes from @Almo7aya (I was working on his branch trying to
understand why some titles wouldn't work for me), plus the Windows 11 24H2
crash fix I found along the way and some build/diagnostic housekeeping.

---

#### Commit per commit

**gpu: bind storage buffers to the mapped extent** (from @Almo7aya)
GNM buffer descriptors routinely over-provision `num_records` (often ~UINT32_MAX),
but the nominal footprint was handed straight to the buffer cache, which tried to
stage/track/map memory the guest never mapped. Depending on where it landed it
caused random-looking access violations or invalid tracker ranges. Now binds the
contiguous mapped extent instead (with `RangeSet::ContiguousExtent` +
`MappedExtent`), clamps when the footprint exceeds `maxStorageBufferRange` instead
of failing hard, and reports once on stdout. Vulkan robust buffer access returns
zeros past the end, which is what the guest's own bounds checks already assume.

**fix(gpu): bind an indexed fetch past its record count** (from @Almo7aya)
Skinned meshes rendered torn: part posed correctly, rest collapsed onto the
origin. The bone-matrix descriptor described fewer records than the mesh
indexed, so bones past the end read zero matrices. Read-only stride-bearing
descriptors now get the mapped extent instead of exactly `stride * num_records`,
capped at 64 KB (a thousand-odd matrices, well past any real mesh, and bounded so
the buffer cache's write-protection doesn't reach unrelated game data).

**fix(gpu): do not let a zero record count bind nothing at all** (from @Almo7aya)
`num_records = 0` hit the empty-footprint shortcut and bound the null buffer, so
the whole mesh collapsed. The indexed read-only fetch is now exempt from that
shortcut and the mapping is asked for instead of taking the descriptor at its
word. Removes the last of the skinned-mesh deformation.

**fix(gpu): keep the indexed-fetch binding on compute too** (from @Almo7aya)
Skinning also happens in a compute skin-cache pass that reads the same bone
matrices, so excluding compute brought the tear straight back. Compute included.

**fix(win): resume guest continuations on Windows 11 24H2** (mine)
On Win11 24H2, `EXCEPTION_CONTINUE_EXECUTION` is dispatched through
`ntdll!RtlGuardRestoreContext`, which validates the continuation RSP against the
thread's TEB stack bounds (`RtlGuardIsValidStackPointer`) and fail-fasts the
process with `0xc0000409 / FAST_FAIL_INVALID_SET_OF_CONTEXT` for any guest-mode
RSP. The fix sets the process shadow-stack policy byte (`LdrSystemDllInitBlock+0x9c`)
so the validation is skipped, redirects the resume through a per-thread
trampoline (host RSP given to the dispatcher, guest RSP swapped back in before
jumping to the guest RIP, without touching any GPR), normalizes `ContextFlags` so
the restore takes the pure user-mode path instead of `NtContinue`, and registers
the dynamic guest ranges as CET-compatible (best-effort, log-only on failure).

**fix: surface unhandled host exceptions with more context** (mine)
Added a `std::terminate` handler that prints a backtrace and exits with a
distinct code instead of the default abort, and on an unresolved access
violation the faulting region's state/protection is logged via `VirtualQuery`
so crashes are easier to diagnose remotely.

**build: add generic Windows build scripts** (mine)
Replaced machine-specific scripts with generic ones that work for anyone:
`configure-windows.bat` finds the MSVC environment via `vswhere` and
auto-detects Qt under `C:\Qt`, `build-windows.bat` builds/installs the launcher,
`fix-submodules.bat` re-checks-out every submodule at the committed SHA.

**chore: ignore emulator runtime data** (mine)
`.gitignore` now covers the download cache, shader cache, misc runtime dirs and
PDB files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,10 @@ build/
 _Build/vscode-clang/
 _Build/
 cmake-build-debug/
+
+# Emulator runtime data
+_DownloadData/
+_Shaders/
+_TempData/
+_Textures/
+*.pdb
--- a/scripts/build-windows.bat
+++ b/scripts/build-windows.bat
@@ -1,0 +1,18 @@
+@echo off
+rem Generic Windows build for KytyPS5. Run scripts\configure-windows.bat first.
+rem
+rem Optional overrides (must match configure):
+rem   BUILD_DIR  output directory, defaults to build
+
+setlocal
+
+if not defined BUILD_DIR set "BUILD_DIR=build"
+
+cd /d "%~dp0.."
+cmake --build "%BUILD_DIR%" --target launcher --parallel %*
+if errorlevel 1 exit /b 1
+cmake --install "%BUILD_DIR%" --prefix "%BUILD_DIR%/install"
+if errorlevel 1 exit /b 1
+
+echo.
+echo Build complete: %BUILD_DIR%\install
--- a/scripts/configure-windows.bat
+++ b/scripts/configure-windows.bat
@@ -1,0 +1,55 @@
+@echo off
+rem Generic Windows configuration for KytyPS5 (clang-cl + Ninja + Qt 6 + Vulkan).
+rem
+rem Requirements:
+rem   - Visual Studio or Build Tools 2017+ with the C++ toolset installed
+rem   - clang-cl, cmake and ninja on PATH (LLVM, CMake, Ninja)
+rem   - Qt 6 built for MSVC x64 (auto-detected under C:\Qt, override with QT_DIR)
+rem   - Vulkan SDK (VULKAN_SDK set by its installer; optional for configure)
+rem
+rem Optional overrides (set before running):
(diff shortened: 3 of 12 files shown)


Pouare514:
broke too many things

Almo7aya:
I have a branch that has the fixes for ur kernel stuff you can check it out here https://github.com/Almo7aya/KytyPS5/tree/windows11-fix_fix 
@Pouare514

Pouare514:
> I have a branch that has the fixes for ur kernel stuff you can check it out here https://github.com/Almo7aya/KytyPS5/tree/windows11-fix_fix @Pouare514

thx  i will take a look at it

PR #237, opened 2026-08-11
Bringup fixes

shadowbeat070:
## Summary

All of these changes were made and discovered when trying to make Beyond Good & Evil 20th Anniversary Edition boot/run. It went from dying on first submit to booting up, starting the game and even rendering/playing the intro cutscene.

- Accept empty DCB submissions instead of aborting.
- Clamp a storage-buffer descriptor to mapped memory.
- Depth compare sampling.
- Emulate SSE4a EXTRQ/INSERTQ register operands.
- C11 plain mutexes now get PTHREAD_MUTEX_NORMAL.
- Texture cache eviction under memory pressure fix.
- Download staging fallback for dowloads bigger than the 32Mib shared stream.
- Split DMA copy rejection into seperate cases and allowed exact self copies.
- Added a fast path for looking up storage buffer mapped ranges from stored
  records. Requests not fully covered by those records fall back to the host
  query as before.
- Added throttling for deferred destroys and device memory in flight, so a
  streaming burst can no longer allocate past the budget while nothing is
  able to be freed.

## Notes for reviewers

- Follows project contributing guidance: focused change, Windows primary build verified.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -563,6 +563,7 @@ if(BUILD_TESTING)
 		resource_tracking_tests
 		event_queue_lifetime_tests
 		sync_on_address_tests
+		audio_out2_port_tests
 		shader_recompiler_compute_tests
 		virtual_memory_allocation_tests
 	)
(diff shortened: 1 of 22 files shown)


Nmzik:
@shadowbeat070 Thanks. Credited to you. Game can actually go in-game with two hacks (I'll fix it)

Nmzik:
@shadowbeat070 I think the game is fully playable now. Please check it on your side. Some of the changes in the PR didn’t need to be merged at all.

PR #244, opened 2026-08-12
launcher: quote emulator paths that contain spaces

### LINKED ISSUE #179: Problem with directories, spaces, or characters (easy solution)
I'm having a problem with the directory where I store the emulator; I keep it on my D: drive, where I store my emulators. Apparently, this emulator has issues with folders containing spaces and fails to run the games.

Example, failing to run games because the "Sony Playstation 5" folder has aspace in its name;

Image

An example of how adding underscores made the emulator work without issues.

Image

However, adding underscores disrupts the order of my emulator folders; please fix this simple issue so the emulator doesn't fail when stored in folders with spaces in their names.

It would also be great if this other topic could be addressed, if possible.  #151

Hashim1999164:
## Summary
- On Linux, run `kyty_run.sh` as a bash file argument instead of `bash -c`, so script paths with spaces still work.
- On Windows, quote the emulator path and arguments and pass them through `setNativeArguments("/K ...")` so cmd.exe does not split install folders that contain spaces.

Fixes #179.

## Test plan
- [ ] Place the emulator under a path with spaces (for example a folder named Sony Playstation 5) and launch a game from the launcher on Windows.
- [ ] On Linux, launch from an install path that contains spaces and confirm the terminal still starts the helper script.
- [ ] Confirm launch still works for paths without spaces on both platforms.

### DIFF
--- a/src/launcher/src/mainDialog.cpp
+++ b/src/launcher/src/mainDialog.cpp
@@ -391,8 +391,8 @@ void MainDialog::RunInterpreter(QProcess* process, const Configuration& info) {
 		// Use nativeArguments so Qt does not re-quote the /K command string.
 		process->setProgram(CMD_EXE);
 		process->setArguments({});
-		process->setNativeArguments(QStringLiteral("/K ") +
-		                            BuildWinCmdKCommand(interpreter, args));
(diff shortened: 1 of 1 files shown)


williansti:
Thanks for that.

PR #250, opened 2026-08-13
gpu: support required render states

Senaxx:
This adds handling for several rendering states currently needed by more demanding PS5 workloads.

The changes improve depth render-target setup, pipeline state selection and draw-state handling. Unsupported combinations remain guarded instead of silently producing an invalid pipeline.

The implementation is generic and does not contain any title-specific checks.

Note:
This is based on the work to make Demon's Souls to boot.


Nmzik:
@Senaxx credited to you

PR #252, opened 2026-08-13
gpu: preserve sparse and mapped guest memory

Senaxx:
This improves the lifetime and synchronization of sparse or partially mapped guest memory.

Buffer cache entries are updated when guest mappings change, while valid mapped regions and their contents are preserved. This avoids stale native resources and incorrect assumptions about contiguous guest allocations.

The behavior is implemented at the memory and buffer-cache level and benefits any workload using sparse GPU mappings.

Testing:
- Built virtual_memory_allocation_tests and shader_recompiler_compute_tests.
- Ran virtual_memory_allocation_tests successfully.
- Helped DS compile more shaders
- Afaik no regression on other titles

Note: 
My commits had stacked dependency's, had AI to help me to rewrite them so they could be merged independent of eachother.

### DIFF
--- a/src/graphics/host_gpu/renderer/cache/bufferCache.cpp
+++ b/src/graphics/host_gpu/renderer/cache/bufferCache.cpp
@@ -52,6 +52,27 @@ void BufferCache::Upload(CommandBuffer& command, Buffer& destination, uint64_t d
 	}
 }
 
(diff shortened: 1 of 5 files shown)


Nmzik:
@Senaxx Which game requires this fix?

PR #253, opened 2026-08-13
gpu: support texture descriptor mip views

Senaxx:
This allows image views to represent the mip range selected by the guest texture descriptor.

Previously, view creation could assume the complete image mip range. That is incorrect when a shader intentionally selects only part of an image's mip chain.

The change keeps view creation generic and validates the selected levels against the underlying image.

Testing:
- Added coverage for texture mip-view selection.


Nmzik:
@Senaxx credited to you

PR #255, opened 2026-08-13
gpu: synchronize buffer and image ownership

Senaxx:
This improves synchronization when the same guest memory is accessed through both buffer and image resources.

When an image becomes the authoritative GPU representation, only the corresponding buffer bytes are discarded. Unrelated dirty bytes belonging to the same native allocation remain preserved.

The patch also enables formatless storage-image access for narrow integer formats where the Vulkan view format cannot be represented directly in SPIR-V.

Testing:
- The new ownership and formatless-storage cases pass.


Nmzik:
formatless storage-image access has been merged. Credited. Another fix is hack. I'm closing PR

PR #256, opened 2026-08-13
shader: use Prospero compute metadata

Senaxx:
This derives compute wave size and related shader metadata from the Prospero dispatch information instead of using the older register interpretation.

In particular, wave32 versus wave64 is taken from the dispatch modifier. The compute local size and workgroup inputs are then propagated through compilation consistently.

This follows the Prospero documentation metadata model and removes an incorrect assumption inherited from older GPU register layouts.

Testing:
- Built and ran shader_cfg_tests successfully.
- Added tests for wave-size decoding and compute metadata propagation.


Nmzik:
@Senaxx merged and credited

PR #257, opened 2026-08-13
shader: experimental support Prospero wave64 execution

Senaxx:
Note: This is a big change and maybe needs more discussion. Also shader is currently compiled twice to determine the runtime selection without restructuring the compiler. This was the minimal solution added to make this work. 

Some PS5 compute shaders are made to run as wave64. This means that 64 shader
threads are expected to work together as one group.

The problem is that many GPUs, especially NVIDIA GPUs through Vulkan, only
support native groups of 32 threads. We cannot simply run the shader as two
separate wave32 groups, because both halves can make different branch
decisions and will not automatically share the same EXEC/VCC masks or
wave-local data.

This PR adds three possible paths for these shaders:

**1. Native wave64**
If the GPU and Vulkan driver support a real 64-thread subgroup, Kyty uses it
directly. This is the most accurate and preferred path. Supported on older AMD GCN Hardware for example. 

**2. Logical wave64**
If the GPU only supports subgroup32, Kyty can run the 64-thread PS5 wave as
two physical groups of 32 threads. It will run in 2 vulkan subgroups and joined by using the sharing state. 

Kyty connects both halves using shared workgroup data and synchronization.
This allows supported shaders to keep their wave64 behaviour, including
EXEC/VCC masks, scalar branch decisions, LDS visibility and several
cross-lane instructions.

**3. Legacy wave32 fallback**
If a shader cannot safely use the logical wave64 implementation, Kyty can
recompile it as two separate wave32 groups.

This is not an exact replacement for wave64. It is a compatibility fallback
for shaders that currently cannot be handled by the logical path. In practice,
this can allow a shader to continue working instead of stopping the emulator.

Before choosing a path, Kyty analyzes the shader and checks which wave
operations and control flow it uses. Logical wave64 is only selected when the
shader can be handled safely. Unsupported cases are kept out of that path
instead of silently compiling them with incorrect behaviour.

The selected wave mode is also included in the shader and pipeline cache
identity. This prevents a shader compiled for one mode from accidentally
reusing a cached version made for another mode.

Testing:
- Built kyty_emulator successfully from this branch alone.
- Built and ran shader_cfg_tests successfully.
- Added policy and logical-wave CFG coverage.

### DIFF
--- a/src/graphics/guest_gpu/command_processor/pm4Handlers.cpp
+++ b/src/graphics/guest_gpu/command_processor/pm4Handlers.cpp
@@ -1367,10 +1367,6 @@ static void HwShSetCsRegister(CommandProcessor& cp, uint32_t cmd_offset, uint32_
 			                         Pm4::COMPUTE_PGM_RSRC1_BULKY_MASK) != 0u;
 			cs_regs.threadgroup_configuration = ((value >> Pm4::COMPUTE_PGM_RSRC1_WGP_MODE_SHIFT) &
 			                                     Pm4::COMPUTE_PGM_RSRC1_WGP_MODE_MASK) != 0u;
-			cs_regs.wave_size                 = (((value >> Pm4::COMPUTE_PGM_RSRC1_W32_EN_SHIFT) &
-			                                      Pm4::COMPUTE_PGM_RSRC1_W32_EN_MASK) != 0u)
-			                                        ? 32u
-			                                        : 64u;
 			break;
 		case Pm4::COMPUTE_PGM_RSRC2:
 			cs_regs.scratch_en     = ((value >> Pm4::COMPUTE_PGM_RSRC2_SCRATCH_EN_SHIFT) &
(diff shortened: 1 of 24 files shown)


Senaxx:
Had a few people test this and worked for them and I received 3 reports of shader compilation failing on shader ```0x208a13c00``` It reports a serious ```ErrorDeviceLost``` Research so far show that the shader compiles successfully and executed twice successfully but upon 3rd execution it crashes. 

The failures are all on Nvidia hardware. One tester had it fail on his Nvidia GPU but successfully booted it on AMD GPU.

gandalfthewhite19890404:
What exact AMD GPUs needed for wave64 support?

Senaxx:
> What exact AMD GPUs needed for wave64 support?

All cards can support wave64, only on GCN, RDNA and Nvidia processes them differently.

Nmzik:
This PR is no longer needed, as it doesn’t allow wave64 shaders to run on a wave32 host. We now have two modes: native wave64 and wave32 emulation using lanes 0–31 and 32–63 (they cant communicate between each other in FS.

PR #258, opened 2026-08-13
gpu: preflight overlapping buffer descriptor ranges

Senaxx:
This fixes stale Vulkan buffer views when multiple guest descriptors refer to overlapping memory ranges. This broke the coefficient data passed between the guest Bink compute shaders, producing corrupted luma and chroma output. This fixed the .bk2 video corruption. 

Creating a larger native buffer can replace an earlier cached allocation. If a view was published before that replacement, one dispatch could use two different VkBuffer objects for overlapping guest memory.

The new preflight resolves all buffer and address ranges before descriptor views are published. Overlapping descriptors therefore refer to the same final native allocation with the correct relative offsets.

Testing:
- Added focused tests for inner/enclosing ranges and reversed acquisition order.
- The new preflight tests pass before the existing unrelated multisample-depth baseline failure.

### DIFF
--- a/src/graphics/host_gpu/renderer/cache/bufferCache.cpp
+++ b/src/graphics/host_gpu/renderer/cache/bufferCache.cpp
@@ -526,6 +526,23 @@ BufferBinding BufferCache::ObtainBuffer(CommandBuffer& command, uint64_t vaddr, 
 	return {owner, owner->Handle(), offset};
 }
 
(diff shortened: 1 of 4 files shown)


Nmzik:
@Senaxx Got an idea, thanks. still cant get the logo to display correctly. Maybe the new SSA shader recompiler has a bug. Needs further investigation.

Nmzik:
Credited

PR #265, opened 2026-08-15
graphics: fix storage stencil writes and sampled depth target layout

shadowbeat070:
Two independent depth/stencil bug fixes. 

## Summary
-  Resolve a storage stencil binding to the stencil plane itself instead of the depth owner and copy the written plane into the depth image stencil aspect after the draw.
- placed a depth target in eGeneral if the same draw samples it so the attachment layout still matches after the binding transition.

bugs were identified on silent hill the short message with vulkan validation enabled

PR follows contribution guidelines and and the windows build is verified


Nmzik:
@shadowbeat070 please include game/ppsa code

PR #266, opened 2026-08-15
graphics: fix copy extents, render area, storage usage, and eviction

shadowbeat070:
Four independent fixes for texture cache and image bugs

## Summary
-  Grant storage usage to block compressed images through BLOCK_TEXEL_VIEW_COMPATIBLE view in the matching uncompressed class.
- Only copy the region that both images both images actually have in CopyImage and CopyImageWithBuffer.
- fix the render area to what the attachment views cover and just the guest extents. A same pitch merge means that a 190 texel target can resolve to a 189 wide image.
- No longer evict gpu authored storage images because the garbage collector would skip downloading and discard the only copy in memory. the next draw would be built from empty guest memory  and cause black textures.

bugs were identified while trying to get bg&e stable

PR follows contribution guidelines and and the windows is build verified


Nmzik:
@shadowbeat070 please include game/ppsa code

PR #267, opened 2026-08-15
graphics: fix an upload-callback deadlock and exception filter scope

shadowbeat070:
Two independent fixes: a deadlock during buffer upload, and an exception filter that armed too early.

## Summary

- Do not drain deferred operations from inside an upload callback. A buffer upload can exhaust the
  staging buffer and wait for the GPU while MemoryTracker holds its region locks across the
  callback, and the deferred operations reclaim buffers and unmark tracker ranges, so draining them
  there re-entered the tracker and deadlocked on non-recursive locks. The wait still frees the
  staging space and the callbacks run at the next drain.
- Classify host exceptions before arming the re-entrancy guard, so the guard is only armed for
  exceptions the filter actually handles rather than for every exception passed straight on to
  EXCEPTION_CONTINUE_SEARCH.

### DIFF
--- a/src/common/hostException.cpp
+++ b/src/common/hostException.cpp
@@ -1,6 +1,7 @@
 #include "common/hostException.h"
 
 #include <atomic>
+#include <cinttypes>
 #include <cstdio>
 #include <cstdlib>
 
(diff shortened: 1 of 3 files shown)


Nmzik:
@shadowbeat070 exception filter is fixed. Credited

PR #268, opened 2026-08-15
graphics: add --gpu-assisted-validation and report fatal VkResults

shadowbeat070:
Two diagnostic improvements for Vulkan failures.

## Summary

- Report the VkResult when a queue submit or fence wait fails. The result code is the whole
  diagnosis, since device lost and out-of-device-memory need opposite investigations, and LOGF
  alone goes to the sink --printf-direction selects, which is Silent by default. Mirrored to stdout
  and flushed so it survives a force-kill, and vkResetFences is now covered too.
- Add a --gpu-assisted-validation flag, separate from --vulkan-validation because it is very slow,
  and folded into it because the implementing layer only loads when validation is on. Requires
  bufferDeviceAddress outright, since the layer otherwise disables itself with only a warning and
  leaves you running a slow build that checks nothing.

### DIFF
--- a/src/common/emulatorConfig.cpp
+++ b/src/common/emulatorConfig.cpp
@@ -93,6 +93,10 @@ bool SpirvDebugPrintfEnabled() {
 	return g_config->spirv_debug_printf_enabled;
 }
 
+bool GpuAssistedValidationEnabled() {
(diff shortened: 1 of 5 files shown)


Nmzik:
@shadowbeat070 Nice PR

PR #269, opened 2026-08-15
graphics: dcc fast clear for color render targets

shadowbeat070:
Implemented the dcc fast clear for color render targets.

PS% games clear render targets by writing a key into dcc metadata instead of writing color. Before this change, the registers were decoded and discarded after so each clear was dropped silently. 

Verified

- ctest 29/29.
- Stray: 18,303 clears applied from 21,265 metadata writes; 69 colour targets.
- Silent Hill the short message: 583 clears applied.
- Dreaming Sarah: locked 60.0 FPS, unchanged (no dcc).


Nmzik:
5d83ce82af3a36b0c9aa99cbf879663c391799ab

PR #285, opened 2026-08-18
shader: fix the Linux and macOS build

StefanosCosta:
## Summary

  Fixes the Linux and macOS build, which has been broken since 2026-08-15. Two independent
  compile errors landed with the shader recompiler work — one is a literal-suffix overload
  ambiguity that only exists on LP64, the other is a `throw` under `-fno-exceptions`. Both
  compile cleanly on Windows, so CI has been green there throughout while Linux and macOS
  failed on every run.

  Windows behaviour is unchanged: one change is a spelling of an integer literal that already
  resolved to the same constructor under LLP64, the other sits in a `consteval` branch that
  emits no code on any platform.

  ## What's included

  **Shader recompiler**

  - `Translate.cpp`: spell the whole-quad mask as `IR::Value(uint64_t {0x1111111111111111})`
    instead of a `0x...ull` literal. `IR::Value` has `explicit` constructors for `bool` through
    `uint64_t`; on LP64 `uint64_t` is `unsigned long`, so an `unsigned long long` literal
    matches none exactly and all five are equal-rank conversions — `error: ambiguous conversion
    for functional-style cast from 'unsigned long long' to 'IR::Value'`. The next line already
    used the correct form. Windows is LLP64, where `uint64_t` *is* `unsigned long long`, so it
    resolved as an exact match and never failed.
  - `OpcodeTable.h`: replace the `throw` in `consteval MakeOpcodeTable` with a call to a
    declared-but-never-defined `[[noreturn]] void InvalidOpcodeTable()`. `-fno-exceptions`
    (`src/utils.cmake:120`, every clang build) makes clang reject `throw` when the expression is
    formed, not when constant evaluation reaches it, so the branch being unreachable did not
    help. A non-`constexpr` call is equally illegal in a constant expression but only when
    actually evaluated, so a malformed table still fails the build — now naming the offending
    table — while a valid one compiles. `static_assert` cannot express this check: the condition
    reads a function parameter, which is not a constant expression inside the body.

  ## Platform impact

  - **Windows** — unchanged. The `Translate.cpp` literal already bound to `Value(uint64_t)`
    under LLP64, so the emitted code is identical. `InvalidOpcodeTable` is never odr-used —
    `nm` shows no reference to it in any decode object or in the linked `kyty_emulator` —
    because an immediate function emits no runtime code. No build flags are touched; Windows
    takes the MSVC branch of `utils.cmake` and never had `-fno-exceptions` in the first place.
  - **macOS** — not built here, so this is reasoning rather than a test result. CI builds macOS
    as x86_64, which is LP64, so the `Translate.cpp` ambiguity applies identically and is fixed.
    Whether Xcode's clang also rejected the `throw` is unconfirmed; that change is harmless
    either way.

  ## Testing

  Linux, clang 18.1.3 (matching the `ubuntu-24.04` CI image), Ninja, Release:

  - Clean configure and build of CI's target list — `launcher`, `page_manager_tests`,
    `memory_tracker_tests`, `audio_out2_port_tests`, `ime_dialog_tests`,
    `virtual_memory_allocation_tests` — 1015/1015, zero errors.
  - `ctest -R '^(audio_out2_port|ime_dialog|page_manager|memory_tracker|virtual_memory_allocation)$'`
    — 5/5 passed.
  - Each fix reverted individually against a real translation unit to confirm the exact error
    returns, then restored.
  - Built with `CMAKE_INTERPROCEDURAL_OPTIMIZATION=OFF`; the local toolchain has no `llvm-ar`,
    so the LTO link path CI uses is not covered by the above.

### DIFF
--- a/src/graphics/shader/recompiler/frontend/decode/OpcodeTable.h
+++ b/src/graphics/shader/recompiler/frontend/decode/OpcodeTable.h
@@ -1,6 +1,7 @@
 #ifndef EMULATOR_INCLUDE_EMULATOR_GRAPHICS_SHADER_RECOMPILER_DECOMPILER_OPCODETABLE_H_
 #define EMULATOR_INCLUDE_EMULATOR_GRAPHICS_SHADER_RECOMPILER_DECOMPILER_OPCODETABLE_H_
 
+#include "common/assert.h"
 #include "graphics/shader/recompiler/frontend/decode/ShaderDecoder.h"
 
 #include <array>
@@ -8,8 +9,6 @@
 #include <cstdint>
 
 namespace Libs::Graphics::ShaderRecompiler::Decoder::Detail {
-
-[[noreturn]] void InvalidOpcodeTable();
 
 struct OpcodeMap {
 	uint32_t encoding = 0;
@@ -29,7 +28,7 @@ consteval auto MakeOpcodeTable(const Entry (&entries)[EntryCount]) {
 	for (size_t i = 0; i < EntryCount; i++) {
 		const auto encoding = entries[i].encoding;
 		if (encoding >= EncodingCount || table.indices[encoding] != 0) {
-			InvalidOpcodeTable();
+			EXIT("invalid opcode table\n");
 		}
 		table.entries[i]        = entries[i];
 		table.indices[encoding] = static_cast<uint16_t>(i + 1);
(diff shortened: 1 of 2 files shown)


Nmzik:
@StefanosCosta Thanks

PR #301, opened 2026-08-20
graphics: enable VK_EXT_memory_priority + pageable_device_local_memory

Hyphaed:
I was reading through the texture cache while working on something related and noticed the renderer already queries `VK_EXT_memory_budget` and scales its GC bands off the real reported budget. That's more care than most projects take. But once memory actually gets tight, the only lever available is `RunGarbageCollector()` deleting resources. The driver itself has no idea which of your allocations are worth keeping, so to it everything looks equally disposable.

That's the gap this fills.

`VK_EXT_memory_priority` lets you tag each allocation 0.0 to 1.0 so the driver knows what to push out first. `VK_EXT_pageable_device_local_memory` is what lets it actually page device-local memory out instead of just returning `VK_ERROR_OUT_OF_DEVICE_MEMORY`. Together they give you a cheaper step *before* the GC has to run: the driver quietly demotes something, rather than you destroying a resource and rebuilding it a few frames later.

It doesn't replace the GC. It takes pressure off it.

### Why it's written this way

**Both extensions go on together or not at all.** Priority on its own only changes which allocation gets the OOM, which isn't much use.

**`pageableDeviceLocalMemory` is a feature bit, not just an extension**, so it goes through `getFeatures2()` and is only requested if the driver actually reports it. Enabling the extension without the feature is invalid usage. If a driver has neither, the `pNext` chain it sees is exactly what it saw before this patch.

**VMA needs `VMA_ALLOCATOR_CREATE_EXT_MEMORY_PRIORITY_BIT`** or it ignores `VmaAllocationCreateInfo::priority` entirely, so the priorities would be dead code without it. Easy one to miss.

**Only images get bumped** (0.8, against VMA's 0.5 default). I went back and forth on buffers and left them alone on purpose. A render target that gets demoted has to be paged back in before the next pass samples it, whereas most buffer content you can re-upload from guest memory you're still holding anyway. But `CreateBuffer()` is the generic path, it also hands out vertex and index buffers that might be every bit as hot, and guessing wrong there costs performance with nothing in the log to show for it. Splitting staging and tiler scratch out and giving those a low priority looks right to me, but I'd want to measure it before claiming so.

### On testing

Both changed translation units compile, and `page_manager_tests` passes.

I initially tried to build with GCC and hit `vk::Extent3D` brace-init ambiguity in files I hadn't touched, which had me chasing the wrong thing for a while. Then I actually read `.github/workflows/build.yml` and saw you build with clang on all three platforms, so GCC was never going to be a meaningful signal. Rebuilt with clang, which is clean.

So: verified under clang, matching CI. I haven't measured a frame-time difference and I'm not claiming one, this is about giving the driver information it didn't have rather than a specific speedup.

### Where this came from

Since it's a first PR from me: I work on a Vulkan layer that oversubscribes VRAM by tiering into host DDR and NVMe behind an inflated heap, and both of these extensions carry real weight there under pressure. Your budget-derived GC bands with the in-loop backoff are a better shape than what I had, so I've been borrowing in the other direction. The priority hint seemed like the obvious thing missing next to them.

Happy to resize or split this if you'd rather have it smaller.

### DIFF
--- a/src/graphics/host_gpu/vma.cpp
+++ b/src/graphics/host_gpu/vma.cpp
@@ -26,6 +26,27 @@
 
 namespace Libs::Graphics {
 
+// Allocation priority for VK_EXT_memory_priority: 0.0 means "demote this
+// first", 1.0 means "demote this last", and VMA uses 0.5 when nothing is
+// specified. It is a hint about relative value, so raising one class is
+// enough to order it against everything left at the default.
+//
+// Only images are raised here, deliberately. A demoted render target has to
+// be paged back in before the next pass that samples it, whereas most buffer
+// content can be re-uploaded from guest memory the emulator still holds. The
+// generic CreateBuffer() path also serves vertex/index buffers that may be
+// just as hot as an image, so lowering it wholesale would be guessing, and a
+// wrong guess here costs performance silently. Finer classification (staging
+// and tiler scratch really are cheap to lose) is a sensible follow-up once
+// someone measures it.
+//
+// This complements TextureCache::RunGarbageCollector() rather than replacing
+// it: the GC still frees things, this just gives the driver a cheaper option
+// one step earlier.
+namespace {
+constexpr float MEMORY_PRIORITY_IMAGE = 0.8F;
+} // namespace
+
 namespace {
 
 struct MemoryStats {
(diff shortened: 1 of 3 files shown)


Hyphaed:
Closing this in favour of a single consolidated PR. I got the build diagnosis wrong here (blamed header shadowing when the real answer is that this project builds with clang, not GCC), and rather than leave a correction sitting on top of a wrong claim I'd rather resubmit cleanly with the full set of changes and a proper verification story. Sorry for the noise.

PR #306, opened 2026-08-24
Implement verified GPU, audio, loader, and POSIX/network compatibility paths

Techx3:
## Summary

- implement verified GPU PM4, shader and cache paths
- add strict unresolved-import diagnostics to the loader
- implement SDL AudioIn capture with a safe silence fallback
- preserve Sysmodule, RUDP and Share state with explicit unsupported errors
- implement the POSIX and libSceNet runtime imports observed in title traces
- add regression coverage and an implementation-gap audit

## Validation

- built `kyty_emulator`, `kernel_file_system_tests` and `shader_recompiler_compute_tests` on Windows
- `KernelFileSystemTests: all cases passed`
- `ShaderRecompilerComputeTests: all cases passed`
- smoke-tested startup of seven local titles with `--strict-unresolved-imports`
- verified that the 12 previously unresolved IL2CPP runtime imports observed in The Messenger now resolve to real implementations

## Known limitations

- this does not connect to official PlayStation Network services
- NP/Commerce/WebApi/Trophy online paths remain offline or explicitly unsupported
- advanced Audio3D/ACM/NGS2 work and 61 ALU opcodes remain pending

SDK-specific behavior is documented in `docs/IMPLEMENTATION_GAPS.md`.


Nmzik:
@Techx3 This PR includes some good fixes and new ABIs. The problem is that The Messenger doesn’t call any of these new ABIs, so we can’t test them.

PR #312, opened 2026-08-24
Fixes to get Demon's Souls booting on Linux

chaitan3:
A series of Linux-specific fixes that let Demon's Souls (PPSA01341) boot and run on
Linux. The game already works on Windows on `main`; every change here is targeted at a
Linux-only failure mode and deliberately does not alter Windows code paths.

Note: All fixes were made with help of AI and verified that there was progress in booting the game after every commit in sequence. The unchanged Windows behavior claim is untested as I don't have Windows machine at the moment. This PR should be reviewed by someone who has deeper knowledge of kyty and linux before merging. 
  
## What's fixes

- **libC: guest `struct tm` overflow** — the guest uses the 9-field Orbis `struct tm`
  (36 bytes, no `tm_gmtoff`/`tm_zone`), while glibc's `std::tm` is 56 bytes.
  `std::mktime` writes the full host struct back into the caller's buffer, overflowing the
  guest allocation and clobbering the stack canary; `std::strftime` reads the uninitialized
  `tm_zone`. The Windows `_mktime64` path only writes the matching 36-byte layout, so this
  only failed on Linux. Fix: copy through a sanitized host `std::tm` and write the
  normalized 9 fields back after `mktime`.

- **fiber: context save jumping to a stack address** — `FiberSaveContext` read the return
  address from `(%rsp)`, but the compiler's frame-pointer prologue (`push %rbp`) shifted
  `rsp`, so `(%rsp)` held the saved `rbp` (a stack address). `FiberRestoreContext` then
  jumped into stack data, causing SIGILL when a fiber completed its first run. Fix: rewrite
  both helpers as naked functions (no prologue) with explicit Win64/SysV argument passing.

- **hostException: private altstack for fault handlers** — the guest stack and the
  flexible-memory GPU buffers share the same address region; the GPU write-watching sweeps
  mprotect that region to `PROT_READ`, so a fault handler running on the guest stack could
  lose write permission on its own pages and the process died with a silent SIGSEGV.
  Fix: run Linux fault handlers on a per-thread private stack installed via `sigaltstack`
  with `SA_ONSTACK` (no-op on Windows/Apple).

- **bufferCache: upload sources through the guest backing** — a CPU-modified page whose
  backing was released by a guest unmap has no host mapping on Linux, so `UploadCopies`
  faulted inside the upload callback and aborted with "memory tracker re-entered from
  upload callback". Fix: read each copy's source through `TryReadBacking` (zeroing unbacked
  copies) instead of the guest virtual mapping, and resolve CPU reads of GPU-dirty pages
  that are not in the registered GPU ranges via the buffer cache.

- **bufferCache: stop corrupting the allocation ledger** — `RunGarbageCollector` overwrote
  the Register/Unregister byte ledger with raw VMA heap usage, so a later `Unregister`
  underflowed after a collection retired a large merged buffer. Fix: sample device usage
  into a local for the collection trigger and keep `m_total_used_memory` as a pure ledger.

- **memoryTracker: atomic protection updates + consistent lock order** — the GPU protection
  sweeps ran `mprotect` under the per-region tracking lock, and that `mprotect` iterates the
  guest-address-space registry under its mutex. A guest thread executing a syscall holds
  that mutex, and when a sweep made its own stack read-only its fault handler blocked on the
  region lock while the sweeping thread waited for the address-space mutex: deadlock.
  Fix: every protection path acquires the guest-address-space mutex before any
  region-tracking lock (made recursive), keeping the updates atomic; `InvalidateRegion`
  scopes the guard so `on_flush` runs with no tracking locks held.

## Why it's safe for Windows

All changes are either guarded to the Linux code paths or are cross-platform correctness
fixes (fiber context save, GC ledger, buffer upload/read fallback, lock-order discipline)
that leave Windows behavior equivalent or strictly more correct.

## Testing

- Demon's Souls (PPSA01341) boots and renders on Linux. Windows build untested.

### DIFF
--- a/src/common/hostException.cpp
+++ b/src/common/hostException.cpp
@@ -92,7 +92,7 @@ static LONG WINAPI ExceptionFilter(PEXCEPTION_POINTERS exception) noexcept {
 	return EXCEPTION_CONTINUE_SEARCH;
 }
 
-void SetupSignalStack() {}
+void SetupSignalStack() noexcept {}
 
 #elif defined(__APPLE__)
 
@@ -165,7 +165,7 @@ static void SignalHandler(int sig, siginfo_t* si, void* uctx) {
 	sigaction(sig, &dfl, nullptr);
 }
 
-void SetupSignalStack() {}
+void SetupSignalStack() noexcept {}
 
 #else
 
--- a/src/common/hostException.cpp
+++ b/src/common/hostException.cpp
@@ -11,6 +11,7 @@
 #else
 #include <csignal>
 #include <initializer_list>
+#include <sys/mman.h>
 #include <ucontext.h> // IWYU pragma: keep
 #include <unistd.h>
 #endif
@@ -91,6 +92,8 @@ static LONG WINAPI ExceptionFilter(PEXCEPTION_POINTERS exception) noexcept {
 	return EXCEPTION_CONTINUE_SEARCH;
 }
 
+void SetupSignalStack() {}
+
 #elif defined(__APPLE__)
 
 static std::atomic<Handler> g_handler {nullptr};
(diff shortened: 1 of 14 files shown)


Nmzik:
Merged into main branch. Credited

PR #313, opened 2026-08-24
graphics: support NGG merged geometry shaders

shadowbeat070:
Compile a merged ES+GS pair as a mesh shader instead of dropping the draw. Includes the wave64 lane-mask handling it needs, plus fixes for SRT constant reads, shadow sampling, stencil plane writes and depth target layouts.

## Summary

- Decode VGT_SHADER_STAGES_EN for if the draw is a merged ES GS pair and concatenates them together.
- I run the wave64 guest program as a 64 invocation workgroup 
- The launch state the hardware would have supplied is synthesized, because a mesh dispatch has no geometry engine: the merged wave info in s3, the vertex and instance indices, and the primitive connectivity.
- Geometry exports are lowered onto the mesh outputs. The packed connectivity dword becomes gl_PrimitiveTriangleIndicesEXT with its null-primitive bit as gl_CullPrimitiveEXT, POS1's layer field becomes a per-primitive gl_Layer, and GS_ALLOC_REQ becomes OpSetMeshOutputsEXT.
- A draw the mesh path cannot take is refused with a named reason and dropped.

This should probably be discussed/tested more before merging.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -431,6 +431,25 @@ add_executable(shader_vertex_metadata_tests EXCLUDE_FROM_ALL
 )
 target_include_directories(shader_vertex_metadata_tests PRIVATE ${inc_headers})
 
+add_executable(shader_stages_en_tests EXCLUDE_FROM_ALL
+	"${KYTY_TESTS_DIR}/ShaderStagesEnTests.cpp"
+)
+target_link_libraries(shader_stages_en_tests fmt::fmt)
+target_include_directories(shader_stages_en_tests PRIVATE ${inc_headers})
+
+add_executable(merged_geometry_tests EXCLUDE_FROM_ALL
+	"${KYTY_TESTS_DIR}/MergedGeometryTests.cpp"
+	"${KYTY_SOURCE_DIR}/graphics/shader/shaderMergedGeometry.cpp"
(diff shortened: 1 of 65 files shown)


Nmzik:
@shadowbeat070 I’m all for using mesh shaders; I have nothing against them.
Yeap, ES/GS NGG -> mesh shader is probably the ideal way

Do they fix graphical bugs in UE games, or at least some of them?

shadowbeat070:
@Nmzik I know that the pr fixes the green frames in stray but I need to see if i can get other UE games (that actually boot) for testing

shadowbeat070:
image


With certain boot fixes for the game pathless, we get this screenshot on the main menu. This is an improvement compared to main since even if port the fixes to main without this pr work, we only get the UI text and a black bg.

shadowbeat070:
image

This pr was tested together with https://github.com/KytyPS5/KytyPS5/pull/325.

@Nmzik as far as i am aware this feature seems to be working. thoughts?

Nmzik:
@shadowbeat070 I’ll definitely merge it after cleaning it up. Thanks for the hard work! Please don’t rebase anything—I’ll take care of it myself. Give me a little more time.

Nmzik:
Credited

PR #315, opened 2026-08-25
graphics: support Standard64KB color target arrays

brandostrong:
## Summary

- Support array-slice views of Standard64KB 2D color render targets.
- Keep the existing single-layer restriction for depth-tiled render targets.
- Validate the full layered backing size and add an upload/readback regression for a five-layer surface.

## Problem

A five-layer Standard64KB color surface selecting array slice 4 was rejected during render-target setup:

```text
unsupported 64KB texture-tiled render-target view: dimension=1 depth=0 layer=4/5
```

The Standard64KB tiling and transfer paths already address array slices sequentially, but the render-target guard and support predicate only accepted a one-layer backing.

## Implementation

- Allow Standard64KB color targets to use nonzero array layers while leaving depth-tile restrictions unchanged.
- Require at least one layer and validate the total backing size across every layer.
- Add a five-layer GPU upload/readback regression that selects layer 4, matching the captured failure.

## Testing

- `Standard64RenderTarget` passed.
- `RenderExecutorColorStandard64Tile` passed.
- Existing `RenderExecutorColorStandard4Tile` regression passed.
- The complete `ShaderRecompilerComputeTests` suite passed.
- In GTA V, the previous layer 4/5 failure was cleared and execution advanced to the Ludendorff sequence with dialogue audio before reaching a separate shader-decoder limitation.

## AI assistance disclosure

AI assistance was used to inspect the failure path, help implement the change, and draft regression coverage. I reviewed the resulting diff, built it locally, ran the tests above, and validated the change in-game.

### DIFF
--- a/src/graphics/host_gpu/renderer/colorRenderTarget.cpp
+++ b/src/graphics/host_gpu/renderer/colorRenderTarget.cpp
@@ -219,8 +219,9 @@ void RenderExecutor::ResolveRenderColorTarget(uint64_t submit_id, CommandBuffer&
 		     rt.attrib.num_fragments, bytes_per_element, rt.cmask.addr, rt.fmask.addr,
 		     rt.dcc_addr.addr);
 	}
-	if ((standard64 || depth_tile) && (rt.attrib3.dimension != 1 || rt.attrib3.depth != 0 ||
-	                                   view.base_layer != 0 || view.image_layers != 1)) {
+	if ((standard64 || depth_tile) &&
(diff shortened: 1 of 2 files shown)


Nmzik:
@brandostrong Are you using GTA V in Fidelity mode? GTA V goes in-game on the latest commit when using Performance mode. Fidelity mode uses ray tracing, which we don’t support.

brandostrong:
Ah good catch, I get past the error this PR fixes in Fidelity mode.

Nmzik:
@brandostrong Thanks! I cleaned up your PR quite a bit and merged it.

PR #325, opened 2026-08-27
The pathless boot/stability fixes

shadowbeat070:
multiple fixes to make the pathless boot and reach the main menu stabily.

## Summary

- Extended the redzon patcher so it covers leaf functions that keep their red zone at [rbp-k] instead of [rsp-k]
- Added diagnostics for host faults 
- InvalidateCpuAliases skipped no longer skip every image with depth_id set which caused stencil association write watchers to not release.
- Added recognition for d-word granular meta data fills.
- Added recognition for fills whose clear code is loaded from a constant buffer.
- The clear codes are no longer gated by a format whitelist which caused them to be thrown away. Now any float/normalized format is valid.


Nmzik:
A lot of rewrites have taken place. The game now reaches the main menu properly. NGG GS is next.

PR #342, opened 2026-08-28
launcher: add 1440p option and improve toolbar contrast

Techx3:
## Summary

- add 2560x1440 to the launcher resolution choices
- give toolbar actions explicit normal, hover, pressed, focus, and disabled states
- keep both changes within the existing enum-driven launcher flow

## Validation

- `cmake --build _Build/launcher-usability --target launcher --parallel 2`
- launched the resulting Windows executable and verified the toolbar visually
- opened Global settings and confirmed the resolution list contains 1280x720, 1920x1080, and 2560x1440
- closed the dialog without saving, so local user settings were not changed

## Scope

Launcher-only usability change; emulator rendering and game compatibility paths are unchanged.

### DIFF
--- a/src/launcher/include/configuration.h
+++ b/src/launcher/include/configuration.h
@@ -54,6 +54,7 @@ public:
 	enum class Resolution {
 		R1280X720,
 		R1920X1080,
+		R2560X1440,
 	};
 	Q_ENUM(Resolution)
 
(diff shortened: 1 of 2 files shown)


Nmzik:
@Techx3 Credited. The UI is deliberately not accepted.

PR #345, opened 2026-08-28
graphics: add structured PM4 submission inspector

Techx3:
## Summary

Add an opt-in structured PM4 submission inspector alongside the existing text command-buffer dump.

When Command Buffer Dump is enabled, Kyty now writes one JSON trace per graphics or async-compute submission. Normal execution is unchanged when dumping is disabled.

## Captured data

- Root packet words and readable nested indirect command buffers.
- Direct and indirect context, shader, and uconfig register writes.
- Persistent register-state counts and hash per queue.
- Queue, frame, source address, packet offsets, payloads, predicates, and opcode names.
- Explicit Known, Partial, Inferred, Unknown, Unsupported, and UnreadableAtCaptureTime classifications.

Known means the packet structure was decoded. It does not claim complete runtime semantics.

## Safety and correctness

- Validate every mapped guest-memory segment before copying indirect data.
- Cap nesting depth, individual snapshots, and total snapshot size.
- Detect already-captured indirect buffers to stop cycles.
- Keep graphics and async-compute state in separate namespaces even when their numeric queue IDs match.
- Verify directory creation and the exact number of JSON bytes written.
- Preserve existing opcode naming through shared read-only helpers.

## Verification

- pm4_inspector_tests build passed with clang-cl.
- pm4_inspector_tests.exe passed.
- CTest pm4_inspector passed (1/1).
- Full kyty_emulator build passed.

Tests cover direct and indirect registers, nested buffers, malformed and unknown packets, state persistence/reset, JSON schema, trace file output, and cross-kind queue isolation.

## Scope

This PR intentionally excludes the later AGC resource-registration layer from the external branch because those ABI additions require separate validation. It also does not overlap the PM4 fuzzer in #302: that PR stress-tests packet walking, while this change produces deterministic semantic traces for real submissions.

The SDK 12 Shader Core ISA corpus confirms command-buffer interaction at a general level but does not document the private packet names used here. Accordingly, unsupported named opcodes are classified as Inferred, not Known.

Original work is retained through the commit co-author trailer.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -488,6 +488,14 @@ add_executable(image_page_table_tests EXCLUDE_FROM_ALL
 target_link_libraries(image_page_table_tests common fmt::fmt)
 target_include_directories(image_page_table_tests PRIVATE ${inc_headers})
 
+add_executable(pm4_inspector_tests EXCLUDE_FROM_ALL
+	"${KYTY_TESTS_DIR}/Pm4InspectorTests.cpp"
+	"${KYTY_SOURCE_DIR}/graphics/guest_gpu/pm4.cpp"
+	"${KYTY_SOURCE_DIR}/graphics/guest_gpu/pm4Inspector.cpp"
+)
+target_link_libraries(pm4_inspector_tests common fmt::fmt nlohmann_json::nlohmann_json)
+target_include_directories(pm4_inspector_tests PRIVATE ${inc_headers})
+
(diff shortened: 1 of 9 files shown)


Nmzik:
We don't need that on the main branch. Keep it private as a debugging helper for yourself. I also deliberately avoid adding one-time debugging stuff to the project because I don't want to complicate the codebase.

PR #352, opened 2026-08-28
audio: accept streaming ATRAC9 RIFF data chunks

Techx3:
## Summary
- accepts an ATRAC9 RIFF data chunk when its header is present but its declared full payload has not arrived yet
- consumes and initializes the complete RIFF header, then reports partial input until an audio packet is available
- continues requiring complete payloads for fmt, fact, and other metadata chunks
- adds a focused test using a 76-byte header whose data chunk declares 64 KiB

## Problem
RIFF data chunk sizes describe the complete audio payload. Streaming callers may submit the header and current packets only. The parser applied the metadata completeness check before identifying data, so it rejected a valid streaming header whenever the declared stream was larger than the current AJM input buffer.

## Safety
Only the data header is exempted from the full-payload check. The parser does not read beyond the current buffer, and the decoder's existing minimum-frame logic returns AJM_RESULT_PARTIAL_INPUT when no complete packet follows.

## Tests
- ajm_atrac9_streaming CTest passed
- control case confirms a truncated fmt chunk remains unconsumed and rejected as partial input
- full kyty_emulator Release build passed with clang-cl

## Scope
Three files: parser behavior, focused test target, and CTest registration. No codec, PCM, AudioOut, or ABI changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -464,6 +464,12 @@ add_executable(audio_out2_port_tests EXCLUDE_FROM_ALL
 target_link_libraries(audio_out2_port_tests common fmt::fmt)
 target_include_directories(audio_out2_port_tests PRIVATE ${inc_headers})
 
+add_executable(ajm_atrac9_streaming_tests EXCLUDE_FROM_ALL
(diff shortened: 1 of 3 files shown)


Nmzik:
credited

PR #354, opened 2026-08-28
graphics: make Vulkan present mode configurable

Techx3:
## Summary

- Add FIFO, Mailbox, FIFO-relaxed, and Immediate modes to the emulator CLI and launcher settings.
- Persist the launcher choice and pass it to the emulator.
- Select a requested mode only when the active Vulkan surface reports it.

## Fallback and compatibility

Unavailable modes fall back to FIFO. Vulkan requires surfaces to support FIFO: https://docs.vulkan.org/refpages/latest/refpages/source/VK_KHR_swapchain.html. A defensive fallback to the first reported mode also avoids passing an unsupported value if a driver is non-conforming.

## Validation

- Built kyty_emulator and launcher with clang-cl Release.
- Launcher UIC generation and XML parsing passed.
- CLI accepted Mailbox with exit code 0 under --help.
- CLI rejected an unknown mode with exit code 1.

## Scope

This PR changes only presentation selection. Pipeline caching and unrelated performance instrumentation from the source commit are intentionally excluded because active PRs already cover those areas.

Adapted from KytyTouche commit a9bac98; GodsWill is credited as co-author.

### DIFF
--- a/src/common/emulatorConfig.cpp
+++ b/src/common/emulatorConfig.cpp
@@ -37,6 +37,10 @@ bool FullscreenEnabled() {
 	return g_config->fullscreen_enabled;
 }
 
+PresentMode GetPresentMode() {
+	return g_config->present_mode;
+}
+
 uint32_t GetVblankFrequency() {
 	return std::clamp(g_config->vblank_frequency, 30u, 360u);
 }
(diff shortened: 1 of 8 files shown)


Nmzik:
@Techx3 Merged

PR #357, opened 2026-08-28
shader: implement BUFFER_ATOMIC_CMPSWAP

Techx3:
## Summary

- decode MUBUF opcode 0x31 as BUFFER_ATOMIC_CMPSWAP
- model its consecutive VDATA operands explicitly: replacement followed by comparator
- lower the operation to bounds-checked SPIR-V OpAtomicCompareExchange
- preserve existing GLC behavior by writing the previous value back only when requested
- register the new atomic as a side-effecting buffer access

## Implementation notes

The compare-and-swap source pair occupies VDATA and VDATA+1, so address operands begin after both values. The translator records that layout explicitly instead of treating the comparator as an address component.

The SPIR-V emitter uses device scope, the same memory-barrier path as the existing buffer atomics, and emits operands in the OpAtomicCompareExchange order: replacement, then comparator.

LLVM exposes the equivalent AMDGPU operation with src followed by cmp: https://mlir.llvm.org/docs/Dialects/AMDGPU/#amdgpuraw_buffer_atomic_cmpswap-amdgpurawbufferatomiccmpswapop

## Validation

- Release build: kyty_emulator
- CTest shader_cfg: passed
- CTest shader_recompiler_compute: passed
- matching case: memory 10 becomes 99 and GLC returns 10
- non-matching case: memory 20 remains unchanged and GLC returns 20
- generated SPIR-V contains OpAtomicCompareExchange

## Scope

This PR intentionally contains only BUFFER_ATOMIC_CMPSWAP support. Other experimental shader changes are reviewed separately.


brandostrong:
I can concur, I independently implemented this as well as a fix for astrobot.

PR #359, opened 2026-08-28
shader: support sampled 16-bit SINT textures

Techx3:
## Summary

Add sampled-image support for the guest R16_SINT and RG8_SINT layouts without introducing signed Vulkan image descriptors into the existing unsigned image pipeline.

## Implementation

- classify k16SInt and k8_8SInt as supported sampled formats
- use R16_UINT as the Vulkan backing only for sampled views
- retain the original guest format as shader-conversion metadata
- specialize these descriptors as unsigned images so the generated SPIR-V type matches the Vulkan view
- restore each signed component with OpBitFieldSExtract
- preserve native R16_SINT / R8G8_SINT surface mapping and the previous validation policy for storage images

RG8_SINT is deliberately packed into one 16-bit unsigned word. The shader extracts and sign-extends its two 8-bit lanes after the image read, so byte layout and guest-visible 32-bit results are preserved.

## Tests

- ImageLoadR16SintSignExtends: verifies


Nmzik:
894bc7de3c2f2bac1364dd22fcf974ab7a5800e1

PR #361, opened 2026-08-28
shader: preserve logical lane IDs for compute wave64

Techx3:
## Summary

Preserve the PS5 logical lane ID for compute wave64 shaders when the Vulkan host executes the guest wave as two wave32 subgroups.

## Problem

LaneId was emitted from SubgroupLocalInvocationId. In the upper host subgroup that value restarts at zero, so V_MBCNT_LO/HI could assign lanes 32-63 the same indices as lanes 0-31.

## Implementation

- for compute wave64, derive the guest lane as LocalInvocationIndex & 63
- keep the subgroup builtin for other stages and wave sizes
- request LocalInvocationIndex only when a compute wave64 program actually contains LaneId
- document why the compute path differs from the generic subgroup path

## Regression

VectorMbcntWave64UsesLogicalLaneId executes 64 invocations and verifies the ordered lane results


Nmzik:
Current upstream implements it correctly

PR #376, opened 2026-08-28
renderer: allow storage aliases for block-compressed images

Techx3:
## Summary
- map 8-byte BC blocks to R32G32_UINT and 16-byte BC blocks to R32G32B32A32_UINT for storage-capability checks
- advertise storage usage only when that size-compatible format exposes storage-image support
- retain the native BC backing/descriptor format; this change enables compatible uncompressed aliases instead of rewriting sampled descriptors
- create a real BC3-to-RGBA32_UINT storage view in the regression test

## Why
Kyty creates block-compressed images with BLOCK_TEXEL_VIEW_COMPATIBLE and EXTENDED_USAGE. Without eStorage on the backing, a later size-compatible uncompressed storage alias cannot create or bind its view even when the device supports that view format.

This isolates and strengthens the BC-storage part of #266.

## Game / PPSA context
- Scars Above — PPSA08597
- Demon's Souls — PPSA01342

## Validation
- real Vulkan BC3 4x4 to R32G32B32A32_UINT 1x1 storage view created successfully
- built: kyty_emulator, shader_recompiler_compute_tests
- passed: texture_cache_image_overlap
- passed: shader_recompiler_compute
- git diff --check

### DIFF
--- a/src/graphics/host_gpu/renderer/image/image.cpp
+++ b/src/graphics/host_gpu/renderer/image/image.cpp
@@ -46,6 +46,14 @@ namespace {
 	return static_cast<bool>(properties.optimalTilingFeatures & feature);
 }
 
+[[nodiscard]] vk::Format BlockStorageViewFormat(Prospero::BufferFormat guest_format) {
+	switch (Prospero::BlockCompressedBytesPerBlock(guest_format)) {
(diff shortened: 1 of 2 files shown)


Nmzik:
credited

PR #380, opened 2026-08-29
shader: implement DS_BPERMUTE_B32

Techx3:
## Summary
- Decode GFX10 DS opcode 0xA1 as DS_BPERMUTE_B32.
- Lower the instruction to typed IR carrying data, byte address, offset, and execution state.
- Emit a Vulkan subgroup shuffle while preserving the 32-lane row inside wave64 and returning zero for inactive source lanes.
- Reject the unsupported GDS form explicitly.

## Tests
- Adds a decoder regression for the captured words [0xda840000, 0x00000214].
- Adds a 32-lane Vulkan compute test where every lane reads lane 3 and obtains 103.
- Clean standalone build from upstream main: shader_recompiler_compute_tests built successfully.
- Isolated test result: shader_recompiler_compute passed 1/1.

## Runtime validation
Scars Above previously stopped while recompiling shader hash 0xeecd55ed5682b835 at PC 0x50c with unsupported DS opcode 0xA1. With this change, that shader compiles and the title continues running for more than one minute while compiling additional shaders.

## Scope
This PR is one commit based on upstream main. It does not include the ELF/SELF validator fix or the separate D16-HI work.

## Technical provenance
The SDK 12 reference confirms the common 64-bit DS encoding and DS bpermute operation. The numeric 0xA1 mapping was verified against the captured shader binary and the working KytyTouche implementation.


Nmzik:
@Techx3 Official RDNA2 says BPERMUTE opcode 179 (0xb3), while 0xa1 is DS_WRITE_B16_D16_HI ???

Techx3:
@Nmzik Yes, you’re right. I don’t know if you’ve already fixed it, but I submitted a fix in PR #397.

PR #383, opened 2026-08-29
shader: support heterogeneous indirect image tables

Techx3:
## Summary

- Allows sampled float and unsigned-integer images in the same indirect image table.
- Keeps per-candidate dimensionality, cube, conversion format, swizzle, layout, and sampler metadata.
- Emits each candidate with its own typed SPIR-V image operation, normalizes results to guest-visible u32x4, then merges them with OpPhi.
- Selects native or point-filtered samplers per runtime candidate.

## Tests

- Adds resource materialization coverage for a mixed float cube / uint 2D table.
- Extends compute SPIR-V coverage with a float 2D / converted packed-uint cube switch and verifies OpSwitch, OpPhi, both image samples, and bit extraction.
- Clean standalone build from upstream main: resource_tracking and shader_recompiler_compute both pass.

## Runtime validation

Scars Above previously failed while compiling compute shader hash 0xb3ee9eea5b92b870 at PC 0x159c with "indirect image table ... has incompatible candidates". With this change, the same workload ran for more than one minute and reached 873 presented frames without that shader failure.

## Scope

This PR is one isolated commit. It intentionally excludes the ELF tail-size fix, DS bpermute work, sampled-SINT support, and NpCommerce stubs.

## Technical basis

SDK 12 documents image resource constants as typed by their T# dimensionality and format. The dynamic heterogeneous-table behavior implemented here is inferred from the captured shader and validated against the working KytyTouche implementation.

### DIFF
--- a/src/graphics/shader/recompiler/backend/spirv/spirvEmitterImage.cpp
+++ b/src/graphics/shader/recompiler/backend/spirv/spirvEmitterImage.cpp
@@ -610,9 +610,9 @@ bool EmitValueImage(ValueEmitContext& ctx, const IR::Inst& inst) {
 			ctx.Fail(inst, "uses depth comparison with a packed integer image");
 			return true;
 		}
-		const auto coord =
-		    CoordF32(ctx, mem, *address, layout.coord, ImageViewCoordinateComponents(view));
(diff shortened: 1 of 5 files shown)


Nmzik:
Credited

PR #399, opened 2026-08-29
shader: implement V_SIN_F16

MehmetCambaz:
Summary

Rebased onto 1a87cfe, which implemented V_COS_F16. This now adds only its sibling, V_SIN_F16 (VOP1 0x60), still missing from the decode tables — a shader using half precision sine fails to decode. Reported against #281.

Follows the shape of V_COS_F16: entries in the VOP1 and VOP3-encoded VOP1 tables, an SDWA rule, and a translator that widens to f32 and back.

The half-cycle results are architecturally exact zero. Evaluating sin(pi) with a rounded F32 PI lands on a half-precision subnormal (0x8001) instead, so the two crossings are selected back to zero — the same correction V_COS_F16 applies at its quarter cycles. An infinite source produces the invalid result rather than a NaN bit pattern.

Checking the fraction rather than the source also covers negatives, since fract(-0.5) = 0.5.

Encoding cross-checked against LLVM's AMDGPU backend:

defm V_SIN_F16 : VOP1_Real_gfx10<0x060>;
defm V_COS_F16 : VOP1_Real_gfx10<0x061>;
Test plan

Windows 11 25H2, clang-cl 20.1.8, RTX 2060. New VectorSinF16 case in ShaderRecompilerComputeTests, executed on the GPU.

 ctest: 35/35.
 Covers sin(0) = 0, sin(0.5 rev) = 0 exactly, sin(0.25 rev) = 1, sin(-0.25 rev) = -1.
 sin(0.5 rev) is the case that matters: it returns 0x8001 without the exactness correction and 0x0000 with it, so the test fails if that handling is removed. Verified by taking the select out and re-running.
 The negative source covers the sign path, and sin(0.25 rev) = 1 confirms the revolutions-to-radians conversion, which would give 0.247 if the source were treated as radians.

I do not have the reporting title, so I cannot say how much further DRAGON QUEST VII gets — this removes one decode failure and nothing more is claimed.

AI use

Written with AI assistance (Claude). It traced the decode and translate paths, mirrored the V_COS_F16 structure, and ran the builds, GPU runs and the removal check above on my machine. I reviewed the diff and confirmed the opcode against the LLVM definitions.

### DIFF
--- a/src/graphics/shader/recompiler/frontend/decode/ShaderDecoder.h
+++ b/src/graphics/shader/recompiler/frontend/decode/ShaderDecoder.h
@@ -179,6 +179,7 @@ enum class Opcode {
 	V_CEIL_F16,
 	V_TRUNC_F16,
 	V_RNDNE_F16,
+	V_SIN_F16,
 	V_COS_F16,
 	V_SIN_F32,
 	V_COS_F32,
--- a/src/graphics/shader/recompiler/frontend/decode/VectorAluOps.cpp
+++ b/src/graphics/shader/recompiler/frontend/decode/VectorAluOps.cpp
@@ -137,6 +137,7 @@ constexpr OpcodeMap VOP1_OPCODE_LIST[] = {
     {0x5cu, Opcode::V_CEIL_F16},
     {0x5du, Opcode::V_TRUNC_F16},
     {0x5eu, Opcode::V_RNDNE_F16},
+    {0x60u, Opcode::V_SIN_F16},
     {0x61u, Opcode::V_COS_F16},
 };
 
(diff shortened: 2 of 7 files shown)


MehmetCambaz:
After this PR you implemented a similar solution :)

Jeffslash666:
I retested Dragon Quest VII, and now it gets past the logo screen and into the game, but there is no sound and some textures are missing; I will update the information regarding the game's issues later. In any case, it was a huge step forward! S2

MehmetCambaz:
Heyy that's some good news! We will get there one step at a time :)

PR #400, opened 2026-08-29
fix(shader): treat all-ones texture address as null binding

### LINKED ISSUE #317: [GAME STATUS]: Neptunia ReVerse [PPSA02721]
### Game title

Neptunia ReVerse

### Game ID / serial

PPSA02721

### KytyPS5 version

KytyPS5-2026-08-25-663b524

### Compatibility status

Logo

_No response_

### Result details

This Game does launch, but crashes outright after 5 frames, sometimes more and less at idea factory logo. So not what it shows on the github page, which shows the Menu Screen.

### Steps to reproduce the result

1. Open KytyPS5
2. Launch The Game
3. Even with tweaking settings or default, same issue

### Expected behavior

The logo animation shouldnt be crashed but goes past of that, so it should be normally goes in the title menu. This used to be on ealier versions (which you can see on the github startpage), so there is a regression there?

### OS

Windows 10 22H2 / Windows 11 25H2

### CPU

Intel Ultra 5 225F / Intel Ultra 7 268V

### GPU

RTX 4060 610.88 / Intel Arc 140V iGPU

### RAM / VRAM

32GB Ram on both devices / 8GB VRAM on RTX 4060 and 18GB VRAM on a 140V GPU

### Log file upload

No Log was created.

### Screenshots/video

_No response_

### Extra notes

Default settings and anything like that. Even with Vulkan Pointer, Shader, same thing.

YuhiAida:
## Summary

Neptunia ReVerse (PPSA02721) crashes at the Idea Factory logo with:

    unsupported texture mip view: base=7 last=1 ... dwords=ffffffff,ffffffff,...

in ResolveTexture() (descriptors.cpp).

Root cause: the PS5 texture descriptor uses 0xFFFFFFFFFF (40-bit all-ones) as the null-texture sentinel, matching the RDNA null-descriptor convention. ShaderTextureResource::IsNull() only checked Base40() == 0, so the all-ones sentinel was treated as a real texture. Mip-chain validation then failed and the emulator aborted.

## Change

IsNull() now returns true for address 0 OR 0xFFFFFFFFFF. All-ones descriptors take the existing safe null-texture path instead of crashing.

## Testing

- Platform: Linux (CachyOS, x86_64).
- Built locally (clang + ninja, Qt6) with the patch applied.
- Neptunia ReVerse (PPSA02721) now boots past the Idea Factory logo and runs (verified stable for 1+ minute).
- Stock builds (unpatched) reproduce the crash at the same point with the same dwords signature.
- Formatting verified with clang-format using the repo style (src/.clang-format).

## Risk

Low. IsNull() has a single call site (ResolveTexture). The change only adds a second accepted null value to an existing early exit. Games that already worked are unaffected; games that crashed with the all-ones descriptor now take the null path.

Fixes #317

### DIFF
--- a/src/graphics/shader/shaderBindings.h
+++ b/src/graphics/shader/shaderBindings.h
@@ -94,7 +94,11 @@ struct ShaderTextureResource {
 	[[nodiscard]] uint64_t Base40() const {
 		return ((fields[0] | (static_cast<uint64_t>(fields[1]) << 32u)) & 0xFFFFFFFFFFu) << 8u;
 	}
-	[[nodiscard]] bool                   IsNull() const { return Base40() == 0; }
+	[[nodiscard]] bool IsNull() const {
+		const uint64_t address =
+		    (fields[0] | (static_cast<uint64_t>(fields[1]) << 32u)) & 0xFFFFFFFFFFu;
+		return address == 0 || address == 0xFFFFFFFFFFu;
+	}
 	[[nodiscard]] Prospero::BufferFormat Format() const {
 		return static_cast<Prospero::BufferFormat>((fields[1] >> 20u) & 0x1FFu);
 	}


YuhiAida:
Hi @Nmzik can you take a look at this one is very small

Nmzik:
@YuhiAida this is a hack that's why i won't accept it. RDNA2 doesn't say anything about 0xFFFFFFFFFF as a null value. Some corruption happens in pm4 packets. Thanks for understanding. When I have some time, I’ll research it and push it to the PR.

Nmzik:
Fixed

PR #405, opened 2026-08-29
kernel: stop sign-extending batch-map protection and type

brandostrong:
## Problem

`KernelBatchMapEntry::protection` and `::type` are `char`, signed on the build platforms, and every use passes them to an `int`. A protection of `0xf3` widens to `0xfffffff3`.

Harmless for mapping, where only bits 0-5 are tested, but `KernelQueryMemoryProtection` returns the stored value straight to the guest — so a guest that maps with `0xf3` reads back `0xfffffff3`. The `LOGF` in `KernelBatchMap2` already casts both fields through `unsigned char`.

## Change

`char` -> `unsigned char`. `sizeof` is unchanged and the existing `static_assert` still holds; the now-redundant cast in `LOGF` is removed.

Found while chasing an unrelated crash, with no game symptom to attribute to it.

## AI assistance
Let me know if this is slop, claude caught and noted this whilst looking into an unrelated problem.


Nmzik:
It is valid bug actually!

PR #406, opened 2026-08-29
graphics: profile the buffer readback path

brandostrong:
More profiling for the readback path.

## Change

Three `KYTY_PROFILER_FUNCTION()` calls and one include


Nmzik:
We don't need that. If you need to profile keep it in your private fork :)

PR #410, opened 2026-08-29
fix(kernel) implement sceKernelGetOperationMode

YuhiAida:
## Summary

Fixes two unresolved-import stubs that blocked SpongeBob SquarePants: Titans of the Tide (PPSA26893) from booting on KytyPS5.

## Changes

1. **`src/libs/agc.cpp` + `agc.h` + `libAgcDriver.cpp`** — add and register `AgcDcbCopyDataGetSize` under NID `b5u0Jzm8TF8`.
2. **`src/libs/libKernel.cpp`** — implement and register `sceKernelGetOperationMode` under NID `NH6xARDOVv8` (returns operation mode 1 = Game).

## How the NIDs were identified

The PS5 NID algorithm is `base64(sha1(name + salt))` with a custom alphabet. We reproduced it and matched both NIDs against the community NID database (`ps5rs`):

- `b5u0Jzm8TF8` = `sceAgcDcbCopyDataGetSize`
- `NH6xARDOVv8` = `sceKernelGetOperationMode`

## Evidence

Before: the game printed both `Unresolved import stub called` lines and aborted.
After: both stubs resolve; the game boots much further — RHI threads spawn, VideoOut queue is created and flips, UE5 loads localization/plugins/asset registries. (A separate, later UE5 RHI crash remains on the RHI submission thread; this PR fixes the two stub blockers only.)

## Notes

- This supersedes an earlier draft mapping (`b5u0Jzm8TF8` → `AgcGetRegisterDefaults2`) which was incorrect — the NID signature did not match.
- Tested on Linux (CachyOS), NVIDIA RTX 3060 Ti, official build 1df53b1 + this patch.

### DIFF
--- a/src/libs/libKernel.cpp
+++ b/src/libs/libKernel.cpp
@@ -1836,6 +1836,15 @@ uint64_t KYTY_SYSV_ABI KernelGetGPI() {
 	return bits;
 }
 
+int KYTY_SYSV_ABI KernelGetOperationMode(int* mode) {
+	PRINT_NAME();
+	if (mode == nullptr) {
+		return -1;
+	}
+	*mode = 1; // SCE_KERNEL_OPERATION_MODE_GAME
+	return 0;
+}
+
 } // namespace LibKernel
 
 namespace LibKernelWriteThrottling {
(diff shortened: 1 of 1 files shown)


src/libs/agc.cpp
```diff
@@ -143,7 +143,7 @@ struct RegisterDefaultsStorage {
 #include <fmt/format.h>
 
 static constexpr uint32_t GRAPHICS_REGISTER_DEFAULTS_MAX_VERSION      = 12;
-static constexpr uint32_t GRAPHICS_REGISTER_DEFAULTS_FALLBACK_VERSION = 11;
+static constexpr uint32_t GRAPHICS_REGISTER_DEFAULTS_FALLBACK_VERSION = 12;
```
Nmzik: ~@YuhiAida Was this change intentional?~

Nmzik: ~Everything looks good except for this line. As I remember, the actual implementation falls back to version 11.~ Never mind—I added AGC version 13.

YuhiAida: hey, cool, I was still working on it because those imports are solved but the game doesn't boot yet

PR #418, opened 2026-08-30
shader: preserve raw VCC bits in S_MOV_B64

nomolao2-cell:
Problem

S_MOV_B64 treated EXEC and VCC exclusively as lane masks whenever either the source or destination used one of those aliases.

The implementation called `WriteMask64(inst.dst, ReadMask(inst.src0))`, which converted the source into a boolean lane mask instead of preserving its original two 32-bit values.

Hades temporarily stores numeric shader data in vcc_lo and vcc_hi` The font fragment shader later uses those values for its signed-distance-field calculations. Losing the original values caused the shader to output transparent pixels, leaving save-slot labels and dialogue text invisible.

Fix

Copy the original register values with `ReadU32Pair` and `WriteU32Pair`.
The logical EXEC or VCC mask state is then updated separately, allowing these aliases to retain their raw scalar values without losing their mask behavior.

Testing

- Windows Release build completed successfully.
- shader_cfg: 1/1 test passed, 0 failures.
- Tested Hades, PPSA03355, version 01.003.000, on an NVIDIA RTX 3070.
- Before the change,  dialogue text were invisible.
image
- After the change, dialogue text render correctly.
image

- No crashes or new rendering errors were observed during testing.
- No regressions have been found in other games tested so far.

AI use disclosure

I used Codex to assist with the investigation. It helped analyze RenderDoc captures and shader dumps, locate the affected Hades fragment shader, trace the transparent output back to the loss of the raw VCC values, and prepare the initial code change.

I reviewed the resulting diff, built the emulator locally, ran the existing shader_cfg test suite, and personally verified the fix in Hades before submitting this pull request.


Nmzik:
@nomolao2-cell Thanks for finding bug and detailed explanation

Nmzik:
Broad fix: fixing all mask-aware B64 opcodes/64 SAVEEXEC variants

PR #419, opened 2026-08-30
shader: zero-extend native 16-bit destination writes

brandostrong:
Screenshot 2026-08-30 122846

Screenshot 2026-08-30 123733

## Problem

#370 changed native 16-bit destinations to preserve the untouched half. On that commit
GTA V draws red wireframe outlines over geometry, black window interiors and a grey
horizon band.

Bisecting the commits merged that day isolates it — everything up to and including #404
renders correctly, adding #370 does not.

Before #370, VOP2 F16 already wrote the low half, since `ApplyDefaultVop2F16Destination`
forced `sdwa_sel = 4`. So the effective change was that VOP1/VOP3 native writes went from
zero-extending the dword to preserving the upper half.

## Change

`Write16Bits` zero-extends for a destination the instruction did not encode as SDWA, and
merges only when a half is explicitly selected. `ApplyDefaultVop2F16Destination` is
restored — without it VOP2 F16 flips from preserve to zero-extend, a regression the other
way.

The rest of #370 stands: `explicit_sdwa_dst`, the relaxed destination gates, clamp on the
inexact conversions, the `V_CVT_PKRTZ_F16_F32` modifiers.

`NativeAndSdwa16BitDestinationWrites` pins the old semantics, so its expectations move.
Index 10 stays `0xaaaa4200` (VOP2 `V_ADD_F16` DPP) and index 6 keeps its `0xf00d` high half
(explicit SDWA, `DST_UNUSED = PRESERVE`), which is what bounds the change. Index 6's low
half moves `0x3c00` -> `0x7c00`: it reads `WORD_1` of a register a native write filled
earlier, now `0.0h`, so `V_RCP_F16` returns `+inf`.

I could not find preserve-vs-zero for GFX10 native 16-bit destinations stated in the RDNA2
ISA guide or the LLVM GFX10 tables, so the evidence here is the game and not the doc.

Claude bisected this and wrote the change. If preserve is actually right for the hardware
and GTA V is leaning on zero-extension for some other reason, some other fix may be better


Nmzik:
@brandostrong Yeap, both RDNA2 docs and LLVM say native 16-bit writes preserve the untouched half... So, bug may be somewhere else

brandostrong:
> @brandostrong Yeap, both RDNA2 docs and LLVM say native 16-bit writes preserve the untouched half... So, bug may be somewhere else

Yeah, I saw that for 5 specific opcodes, in 12.12 of the rdna2 docs. My only argument is that it says it for those but not necessarily the rule for others, but I wouldn't expect that kind of ambiguity either. This is pretty beyond me in low-levelness though, so I defer to you.

I'll try digging into it a bit deeper.

brandostrong:
I sent myself down a red herring, #370 did not cause the wireframes. I was able to confirm multiple ways that the preserve does in fact happen in situ, using prosperos among other things, totally refuted this PR multiple ways. The red wireframes were separate problem unrelated to this upstream entirely.

PR #423, opened 2026-08-31
renderer: synchronize linear storage images before CPU access

Techx3:
## Summary

Publish GPU-written linear storage images before the guest CPU reads or partially overwrites them. This fixes the missing image-to-CPU visibility observed during Bendy's ambient-probe calculation: the GPU image contained finite pixels while the CPU used stale bytes and produced non-finite lighting coefficients.

- Track independent image read owners on shared host pages, retaining existing buffer/write ownership.
- Route demand readback through the GPU command lane and wait for both GPU completion and deferred backing publication before releasing read protection.
- Preserve padding, disjoint GPU buffers, partial CPU writes and later image updates; reject ambiguous GPU image ownership explicitly.
- Use non-faulting buffer upload reads while pages are protected. Direct/PRT backing stays first; registered private code/runtime/stack memory uses bounded OS reads. This includes the fix for the Demon’s Souls startup regression caused by requiring every upload to have a direct backing alias.

The path is limited to registered, linear, uncompressed, non-depth, single-subresource storage images with a supported download plan fitting the existing download buffer. The optional eager linear-image readback remains off. No lighting constants, shaders or NaN replacement are introduced.

## Validation

- Release emulator and test executables built independently on Windows with clang-cl/Ninja from `a9166f8`.
- 12/12 focused CTests passed, including both new Vulkan tests, PageManager, MemoryTracker, BitArray, image views, storage sampling, buffer-cache GC, scheduler, GPU command lane and stream buffer. The existing depth-readback CTest exercises host-side footprint checks on this Windows build; it is not counted as GPU depth validation.
- New Vulkan regressions verify exact image publication, shared pages, padding, disjoint dirty buffers, partial CPU writes, repeated ownership and unmap. Private-upload coverage includes protected aliases, private/unmapped rejection and both staging/temporary uploads. PageManager covers shared read ownership and overflow.
- Repeated both new Vulkan tests with `VK_LAYER_KHRONOS_validation`; passed with no reported validation errors.
- Earlier combined local builds verified finite ambient-probe pixels and all 54 source/derived lighting coefficients, restored visible ambient illumination, and progress past the reported Demon’s Souls upload failure. These are integration results, not a full-game compatibility claim.
- An isolated Bendy run of this branch stops before the menu at the unchanged upstream `depthRenderTarget.cpp` check: `stencil state without an active stencil attachment`. Its separate local prerequisite is excluded; standalone gameplay on this upstream base is not yet validated.

## Scope and limits

Independent branch from official `main`, with no dependency on the separate DCC clear PR. No Inspector, diagnostic tooling, raw game/SDK artifacts, HDR display presentation, FSR, NpCommerce or unrelated shader changes. Windows is the locally validated platform; Linux/macOS OS-read branches require CI/platform validation. Full-game compatibility, including unrelated Demon’s Souls shader failures, is not claimed.

### Observed regression

User-supplied screenshot of the affected local build: most corridor surfaces are nearly black and only the distant light remains visible.

Bendy: observed loss of ambient illumination

### Expected appearance — reference

Gameplay reference showing the ambient illumination after the PR.

Bendy: user-provided reference for expected ambient illumination

### Verified result in the current local integration

- Bendy and the Ink Machine, PPSA27616 v01.000.002, 720p, Native filtering; optional eager linear-image readback disabled.
- Before integration: all 54 source/derived ambient-lighting coefficients were invalid (48 NaNs and 6 negative infinities). After integration: all 54 were finite and nonzero ambient lighting returned. Both coefficient arrays exactly matched the previously accepted reference run in two samples.
- Gameplay inspection confirmed visible, illuminated wooden surfaces and the ink-machine room. Those observations were at different camera positions from the supplied corridor images; a matched-camera screenshot A/B is not included here.
- Combined local Windows build: 24 focused CTest suites passed, 317 shader cases passed, and the dedicated image-to-CPU readback regression passed with Vulkan validation enabled and no reported validation errors.

These new results validate the correction adapted into the current combined local build. They do not replace the original isolated-branch validation above or establish that the unchanged PR head was retested on its own. No full-game compatibility or measured FPS improvement is claimed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -557,6 +557,10 @@ if(BUILD_TESTING)
 		COMMAND $<TARGET_FILE:virtual_memory_allocation_tests> --red-zone-patcher-only)
 	add_test(NAME command_scheduler_timeline
 		COMMAND $<TARGET_FILE:shader_recompiler_compute_tests> --scheduler-only)
+	add_test(NAME shader_recompiler_image_cpu_readback
+		COMMAND $<TARGET_FILE:shader_recompiler_compute_tests> --image-cpu-readback-only)
+	add_test(NAME buffer_cache_private_upload
+		COMMAND $<TARGET_FILE:shader_recompiler_compute_tests> --private-buffer-upload-only)
 	add_test(NAME stream_buffer_ring
 		COMMAND $<TARGET_FILE:shader_recompiler_compute_tests> --stream-buffer-only)
 	add_test(NAME gpu_command_lane
--- a/src/graphics/host_gpu/pageManager.cpp
+++ b/src/graphics/host_gpu/pageManager.cpp
@@ -128,8 +128,9 @@ uint64_t PageEnd(uint64_t vaddr, uint64_t size) {
 
 struct PageManager::Impl {
 	struct PageState {
-		uint8_t write_watchers  : 7 = 0;
-		uint8_t access_watchers : 1 = 0;
+		uint8_t write_watchers = 0;
+		// Buffers and disjoint images may independently own bytes on the same host page.
+		uint8_t access_watchers = 0;
 
 		[[nodiscard]] uint32_t Perms() const noexcept {
 			if (access_watchers != 0) {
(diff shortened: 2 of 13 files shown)


Techx3:
I see that the error is present in main, add the capture to help you see the incident. @Nmzik

PR #427, opened 2026-08-31
Boot Marvel's Spider-Man: Miles Morales through the intro movie

itsmemac:
## What this does

Takes Marvel's Spider-Man: Miles Morales from failing on its third frame to playing the full intro movie, on a stock Windows build with no environment variables set.

Grouped into 18 commits so they can be cherry-picked independently. Each is described below with why it exists. **Commits 1 and 11 are the ones to look at first** — 1 is an unrelated build fix that stands alone, and 11 is the only one that changes behaviour for every title.

## Testing

- Windows, clang-cl, MSVC 14.29 (VS2019) STL, NVIDIA RTX 3060 Laptop (6 GB).
- `ctest`: 33/35. `shader_recompiler_compute` and `texture_cache_image_overlap` fail — **they fail identically on unmodified `main`**, verified by building upstream with only commit 1 applied. This PR introduces no test regressions.
- Spider-Man: reaches the end of the intro (~145 s, ~350 flips), then stops at  `masterSemaphore.cpp:56`. The menu does not come up yet.
- Astro's Playroom: stops on unmodified `main` at a shader CFG failure after 3-5 s; with this PR it gets past that and reaches a different, later failure in PM4 register decode (`unknown sh reg at 01169: 0xca`). Still not playable, but not a regression.
- Not tested on Linux or macOS, and not tested on any other title.

## The commits

**1. `build: fix clang-cl compile of ResourceTracking`** — `main` does not compile with clang-cl today. `Tracker::IndirectImagePlan` is in an anonymous namespace, so `std::projected` cannot be formed and the three `std::ranges`
algorithms over it are rejected. CI misses it because it never builds `kyty_emulator`. Independent of everything else here.

**2-5. Self-contained shader fixes** — `V_CMP_LT_U64` was missing from the VOPC table; wave64 subgroup broadcast needs workgroup-wide lowering because `OpGroupNonUniformShuffle` at Subgroup scope only reaches 32 lanes (DS_BPERMUTE
and DS_SWIZZLE deliberately keep the raw subgroup id, being defined on independent halves); `S_CMP` against an exec mask is the "are any lanes live" idiom and needs a real ballot; buffer loads were clobbering lanes the exec mask had masked off.

**6-8. Descriptor and control-flow resolution** — typed phis and `ReadFirstLane` in the SRT walker (the latter only for non-image descriptors, where every lane holds the same handle); loops whose selection merge escapes the body; `S_SWAPPC_B64` indirect calls resolved from the dispatch side, where the user data is already in hand, which is the only place that breaks the circular dependency described in the commit message.

**9. `recompiler: allow a shader to be refused instead of aborting`** - the change that actually unblocks the boot. `CompileOptions.non_fatal` / `CompileResult.unsupported`; the three give-up points record and unwind instead of `EXIT`. Nothing changes when `non_fatal` is clear.

**10. Tessellation patch control points** draws were hardcoded to 3 control points; take the real count from `VGT_LS_HS_CONFIG.HS_NUM_INPUT_CP` and key the pipeline cache on it.

**11. `graphics: size the buffer and page-table constants explicitly`** - **this one changes behaviour for every title and is the one I would question.**
`DOWNLOAD_BUFFER_SIZE` 32 -> 128 MiB, because this title asks for a single 80 MiB download allocation; and `CACHING_ADDRESS_BITS` 39 where the table previously covered 40 bits, halving the BDA page table to 256 MiB at the cost of
covering guest addresses only to 512 GiB. Both are tuned against one title. If the 1 TiB window matters elsewhere, take the naming and leave the value at 40.

**12-13. Device-memory accounting and tiler scratch** — the caches were deciding when to collect from the driver's process-wide estimate, which varied by ~4x between reads of the same moment here; read VMA's own accounting instead.
Tiler scratch was unbounded across a burst of detiles: measured 409 MiB across 40 live buffers while an image allocation was failing 414 MiB short.

**14-18. Remaining fixes** — image/tiling/video-out, decode and emitter corrections, kernel memory and event-flag paths, runtime linker, and test updates.

## Known gaps

- The menu still does not appear; this gets to the end of the intro only.
- No test covers the non-fatal mechanism in commit 9.
- Single platform, two titles.


Nmzik:
Closing this in favor of your #780.

PR #430, opened 2026-08-31
shader: flush a NaN clamp result to zero

brandostrong:
## Problem

The `CLAMP` output modifier lowers to GLSL `FClamp`, which is undefined for NaN. The hardware flushes a NaN result to +0 when `MODE.DX10_CLAMP` is set, and `shader.cpp` already rejects any shader with `DX10_CLAMP` clear (`EXIT_NOT_IMPLEMENTED(rsrc1.dx10_clamp != true)`), so flushing is the only correct behaviour for every shader the recompiler accepts. On current hardware `FClamp` happens to return 0 for NaN, so this is latent rather than visible — the emitted module just does not say what it means.

See the RDNA2 ISA guide, `DX10_CLAMP`: "Used by the vector ALU to force DX10-style treatment of NaNs: when set, clamp NaN to zero; otherwise, pass NaN through."

## Change

Select the NaN case explicitly in `FPSaturate32` rather than relying on `FClamp`, matching the treatment `V_MIN`/`V_MAX` already get in `EmitMinMaxF32Value`.

`VectorVop3ClampFlushesNanToZero` compiles a `V_ADD_F32` with `CLAMP` set over a NaN source and requires the NaN test to appear in the emitted SPIR-V. It fails on the previous emitter with `missing required text: 2139095040` — the NaN exponent pattern that only the explicit test introduces.

## AI assistance

Confirmed the DX10_CLAMP semantics against the RDNA2 ISA guide, found that the emulator already requires the bit to be set, wrote the change and the regression test.

### DIFF
--- a/src/graphics/shader/recompiler/backend/spirv/spirvEmitterAlu.cpp
+++ b/src/graphics/shader/recompiler/backend/spirv/spirvEmitterAlu.cpp
@@ -405,11 +405,17 @@ bool EmitValueAlu(ValueEmitContext& ctx, const IR::Inst& inst) {
 		case IR::ValueOpcode::FPNeg32:
 			ctx.Define(inst, EmitFNegateValue(state, ctx.Arg(inst, 0)));
 			return true;
-		case IR::ValueOpcode::FPSaturate32:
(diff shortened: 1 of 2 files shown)


Nmzik:
PR duplicate

PR #434, opened 2026-08-31
libSaveData: persist save metadata so saved games can be loaded

CheesyPoofs346:
## Summary

- Persist save metadata (title, subtitle, detail, mtime) so a title can find and load its own saves.
- `SaveDataSetParam` discarded what it was given, `SaveDataGetParam` always returned zeros, and `SaveDataDirNameSearch` blanked the params of every save it listed.

## Problem

`SaveDataSetParam` logged the parameters and returned `OK` without storing them:

```cpp
if (param_type == 0) {
    const auto* p = static_cast<const SaveDataParam*>(param_buf);
    LOGF("\t title      = %s\n" ... , p->title, p->sub_title, p->detail, p->user_param);
}
return OK;
```

`SaveDataGetParam` zeroed its output regardless of what was on disk:

```cpp
std::memset(param_buf, 0, sizeof(SaveDataParam));
return OK;
```

And `SaveDataDirNameSearch` found the save directories correctly but cleared their params:

```cpp
if (result->params != nullptr) {
    result->params[i] = {};
}
```

So the metadata round-trip was entirely missing. A game enumerating its saves received a valid directory name with an empty title and `mtime = 0`, which is indistinguishable from an empty slot.

Observed in Grand Theft Auto V (PPSA04264): the game wrote a valid 565 KB save to `SAVEDATASGTA50000` and mounted it successfully, but on the next launch the load menu did not offer it and starting the game began a new playthrough. The save data itself was intact the whole time — only its description was lost.

This change stores the `SaveDataParam` next to the save data as `sce_param.bin`, reads it back in `SaveDataGetParam`, and fills it in for each entry returned by `SaveDataDirNameSearch`. `mtime` is set from the wall clock when a title does not supply one. Saves written before this change have no metadata file; the search still lists them, and they gain metadata on the next save.

`SaveDataMountSlots` gains a `Directory()` accessor so a mount point can be resolved back to the directory its metadata belongs in.

## Validation

- Built on Windows with clang-cl/Ninja; the existing test executable passes.
- Verified in GTA5: with this change the save is offered in the load menu and loads to the correct position, where previously the same save data started a new game.

## AI assistance

### DIFF
--- a/src/libs/libSaveData.cpp
+++ b/src/libs/libSaveData.cpp
@@ -16,6 +16,8 @@
 #include <cstring>
 #include <deque>
 #include <vector>
+#include <ctime>
+#include <fstream>
 
 namespace Libs {
 
@@ -319,7 +321,8 @@ int KYTY_SYSV_ABI SaveDataTerminate() {
 
 	Common::LockGuard lock(g_mount_mutex);
 	if (!g_mount_slots.Empty()) {
-		return SAVE_DATA_ERROR_BUSY;
+		::printf("SaveDataMount3: refused (%s)\n", "SAVE_DATA_ERROR_BUSY");
+	return SAVE_DATA_ERROR_BUSY;
 	}
 	g_save_data_events.clear();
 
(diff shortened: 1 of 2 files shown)


Nmzik:
Credited

PR #440, opened 2026-09-01
renderer: materialize DCC clears before sampled access

brandostrong:
This is the whiteout/hdr fix that was affecting multiple games

## Problem

A DCC fast clear is materialized only as a render-pass load-op, at the next time the surface is
bound as a colour render target. If the guest **samples** that surface as a texture before that
bind, the read returns the pre-clear contents.

I hit this in GTA V. An ordered, per-draw trace of one frame shows a full-screen blit sampling a
1280x720 `B10G11R11` transient at draw 101, while the surface's first render-target bind — and so
its clear — is draw 103:

```
seq=101  reads  0x25EF80000, writes the HDR scene
seq=103  writes 0x25EF80000   <- first render-target bind this frame, clear applied here
seq=104+ writes 0x25EF80000   <- bloom threshold, separable blur, additive upscale
```

The surface is reused several times per frame, and at draw 101 it still held the previous frame's
bloom. The blit adds it into the HDR scene with `ONE/ONE`, the next frame's bloom threshold reads
that scene, and the loop gains roughly 1.7-2x per frame until the image saturates. Auto-exposure
compensates until it hits its lower clamp and then cannot keep up.

## Why

A fast clear is a metadata write that takes effect when it is issued. Any later read of the
surface — as a texture or as a render target — must observe cleared data. Deferring to the next
render-target bind is only valid if nothing reads the surface in between.

The same requirement is visible in AMD's open-source PAL driver: a colour surface that will be
texture-fetched needs a fast-clear-eliminate pass before it can be sampled, and only a small set
of hardware-defined clear colours avoid it
(`src/core/hw/gfxip/gfx9/gfx9MaskRam.cpp`, `Gfx9Dcc::GetFastClearCode` and `pNeedFastClearElim`).

## Change

- Move the DCC clear-code decode out of the draw path into `TextureCache`, so both the
  render-target path and the sampled path resolve it the same way.
- `FindTexture` now materializes a pending DCC clear with `clearColorImage` before the image is
  used, then consumes it. The render-target path keeps the cheaper load-op.
- Carry the target's clear value and support flags on `ImageDesc` so the cache can resolve a clear
  without the draw-time register state.
- Report an unrecognised DCC clear code once instead of silently skipping the clear.

No title, address or shader-hash special cases.

## Testing

- `kernel_file_system_tests`, `resource_tracking_tests`, `scalar_provenance_tests`,
  `shader_cfg_tests`, `virtual_memory_allocation_tests` pass.
- `shader_recompiler_compute_tests` fails at `RenderExecutorStencilBindingDiscovery`
  ("depth texture compatibility identity"). This reproduces unchanged on `main` at `dd062ce`
  with no patch applied, so it is pre-existing and unrelated; the existing
  `UnifiedTextureCacheFlow` case was extended to assert that a pending clear is materialized on
  the sampled path.
- GTA V: the saturation is gone and bloom renders normally. Scene mean 0.041-0.076 with no
  non-finite values, against a saturated 48099 before.

## Known gap

GFX10's comp-to-single clear code (`0x10`) is still unhandled — its colour lives in the first
pixel of each DCC block rather than in a register, so it needs a different value source. It does
not occur in the title I tested; the new report makes it visible rather than silent if it does.

## AI involvement

Investigation, patch and testing were done with AI assistance

### DIFF
--- a/src/graphics/host_gpu/renderer/cache/textureCache.cpp
+++ b/src/graphics/host_gpu/renderer/cache/textureCache.cpp
@@ -1250,6 +1250,7 @@ vk::ImageView TextureCache::FindTexture(ImageId id, const ImageDesc& desc) {
 	if (!image.info.data.Empty()) {
 		RefreshImage(id, desc);
 	}
+	MaterializeDccMetaClear(id, image, desc);
 	switch (desc.type) {
 		case BindingType::Texture: break;
 		case BindingType::Storage:
@@ -1293,6 +1294,9 @@ vk::ImageView TextureCache::FindRenderTarget(ImageId id, const ImageDesc& desc) 
 		} else if (!inserted && metadata->second.type != MetaDataInfo::Type::Dcc) {
 			EXIT("TextureCache: color target reuses non-DCC metadata\n");
 		}
+		metadata->second.color_clear_value           = desc.metadata_clear_value;
+		metadata->second.color_clear_supported       = desc.metadata_clear_supported;
+		metadata->second.fixed_color_clear_supported = desc.metadata_fixed_clear_supported;
 	}
 	CommitGpuWrite(image);
 	TrackImageDownload(id, image);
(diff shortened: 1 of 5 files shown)


brandostrong:
0x10 is yet to be supported, I'll add it in a further pr. I do not have any games that use it though, so is it better to leave it off as a crash/error to fix with a game as backing, or just to try to implement? 

correction: gta 5 does fire some 0x10 clears

## DCC fast-clear codes

A DCC fast clear writes only the metadata byte for each block — no pixels are touched. The code
records *how a reader recovers the colour*.

| Code | Name | Colour | Colour stored in | On clear (write) | On read (sample) | Eliminate needed? | Kyty |
|---|---|---|---|---|---|---|---|
| `0x00` | transparent black | `(0, 0, 0, 0)` | implied by the code | stamp metadata byte only | texture unit returns the implied colour | No | supported|
| `0x40` | opaque black | `(0, 0, 0, 1)` | implied by the code | stamp metadata byte only | texture unit returns the implied colour | No | supported|
| `0x80` | white, zero alpha | `(1, 1, 1, 0)` | implied by the code | stamp metadata byte only | texture unit returns the implied colour | No | supported|
| `0xC0` | opaque white | `(1, 1, 1, 1)` | implied by the code | stamp metadata byte only | texture unit returns the implied colour | No | supported|
| `0x20` | comp-to-reg | any | `CB_COLOR*_CLEAR_WORD0/1` | stamp metadata byte; colour stays in registers | texture unit cannot see those registers | **Yes** | supported|
| `0x10` | comp-to-single (GFX10) | any | first pixel of each DCC block, in image data | stamp metadata byte; write colour to pixel 0 of each block | texture unit reads pixel 0 of the block | No | **not handled** |

PR #449, opened 2026-09-02
Fix GTA V boot-time text glyph loss

CheesyPoofs346:
## Summary

- Fix GTA V boot-time text glyph loss by materializing the boot font's BDA-backed source pages before its one-shot compute dispatch.
- Keep incompatible storage and sampled R8 atlas views separate, publishing storage-written atlas bytes before the sampled view is created.
- Add focused coverage for BDA range prefetch and the 2048x4096 R8 Standard64KB text-atlas layout.

## Problem

In Grand Theft Auto V (PPSA04264), the boot/legal screen and the UI shown before gameplay rendered only scattered glyphs. Characters remained at their correct positions, but most glyphs were invisible; for example, `Performance` appeared as `f  manc`. The in-game settings screen could render correctly, so this was not a missing-font-file problem.

The same font atlas was available to other text draws in the same frame, which ruled out missing glyph assets. The surviving characters also preserved their original spacing, ruling out displaced geometry. The failure was specific to the boot-time rendering path.

## Cause

The boot font compute shader reads its glyph source through buffer device address (BDA). During the one-shot boot atlas build, dynamically referenced source pages could be read before they had been materialized in the buffer cache. Those reads returned unavailable or zero data, leaving the affected glyphs invisible.

A second cache issue could merge the atlas storage image and its differently formatted sampled view. The text atlas uses an R8 integer storage view and an R8 normalized sampled view; treating those as one image owner could leave the sampled view with stale or incorrectly interpreted contents.

## Fix

- Add `PrepareBdaRange`, which page-aligns and materializes only mapped pages in a bounded range before synchronizing BDA resources.
- Apply that prefetch to the identified GTA V boot-font compute dispatch, deriving the source address from the dispatch user data.
- Prevent incompatible R8 storage/sampled atlas owners from being merged.
- Publish GPU-written storage-image contents before separating the cache owners.

The prefetch is bounded and targeted to the identified boot-font dispatch; normal DMA dispatches retain the existing BDA preparation path.

## Verification

- Built with clang-cl/Ninja:
  - `ninja -C _Build/profile-buffer-readback shader_recompiler_compute_tests`
  - `ninja -C _Build/profile-buffer-readback kyty_emulator`
- Focused regression test passed:
  - `shader_recompiler_compute_tests.exe --bda-prefetch-only`
- Launched GTA V through the save-backed test launcher; save slots were detected and backed up before launch.
- Verified in GTA V that the previously missing boot/start text renders, while the in-game settings text remains correct.

### DIFF
--- a/src/graphics/host_gpu/renderer/cache/gpuResourceManager.cpp
+++ b/src/graphics/host_gpu/renderer/cache/gpuResourceManager.cpp
@@ -3,6 +3,9 @@
 #include "common/assert.h"
 #include "graphics/guest_gpu/graphicsRun.h"
 #include "graphics/host_gpu/renderer/commandScheduler.h"
+
+#include <limits>
+
 namespace Libs::Graphics {
 
 GpuResourceManager::GpuResourceManager(GraphicContext& graphics, CommandScheduler& scheduler)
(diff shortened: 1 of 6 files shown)


CheesyPoofs346:
Closing this GTA V-specific version; a generalized BDA prefetch implementation will replace it.

PR #450, opened 2026-09-02
Generalize BDA prefetch and storage/sampled image handling

CheesyPoofs346:
## Summary

- Prefetch mapped BDA pointer neighborhoods from runtime user data and flattened SRT bindings for DMA shaders.
- Keep storage and sampled image cache owners separate when incompatible integer/normalized formats overlap.
- Publish GPU-written image data before separating cache owners.
- Add focused coverage for generic BDA pointer discovery and storage/sampled image separation.

## Problem

Some compute dispatches can read guest addresses before the corresponding pages have been materialized in the host buffer cache. This can produce missing or zero data, including incomplete text or glyphs. Separately, merging an R8 integer image with an R8 normalized view can leave the sampled view using the wrong cache owner or stale contents.

## Implementation

The dispatch path now scans adjacent 32-bit words in user-data and flattened SRT bindings for mapped 64-bit guest pointers, then prefetches a bounded surrounding range. It has no title-specific shader address or game-specific condition, and retains the existing full mapped-range fallback when no pointer is found. Image cache overlap handling now separates incompatible storage/sampled R8 owners and publishes pending GPU writes before reinterpretation.

## Verification

- git diff --check passes.
- The changed renderer translation units compile in the isolated Windows clang-cl/Ninja build.
- The full focused test target was attempted; the first run exposed a stale cache API reference, which was corrected before this commit. A subsequent clean target build was stopped before completion at the user's request.

### DIFF
--- a/src/graphics/host_gpu/renderer/cache/gpuResourceManager.cpp
+++ b/src/graphics/host_gpu/renderer/cache/gpuResourceManager.cpp
@@ -3,6 +3,9 @@
 #include "common/assert.h"
 #include "graphics/guest_gpu/graphicsRun.h"
 #include "graphics/host_gpu/renderer/commandScheduler.h"
+
+#include <limits>
+
 namespace Libs::Graphics {
 
 GpuResourceManager::GpuResourceManager(GraphicContext& graphics, CommandScheduler& scheduler)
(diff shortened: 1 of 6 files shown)


Nmzik:
Glue on top of bug.

PR #464, opened 2026-09-02
renderer: bind vertex buffers with the size of their descriptor

brandostrong:
**Problem.** Vertex buffers were bound with a size derived from the draw rather than from the V#. A fetch past `num_records` then read whatever memory followed the buffer, where the hardware returns zero for out-of-range records (buffer descriptor clamping, RDNA2 ISA "buffer resource" section).

**Change.** Bind the vertex buffer with the descriptor's own size so the host's bounds behaviour lines up with the guest's.

**Effect.** No isolated measured change; part of the GTA V rendering bring-up where out-of-range vertex fetches showed up as garbage attributes.

**Verification.** Suites pass.

**References**
- AMD RDNA 2 Instruction Set Architecture Reference Guide, §8.1.5 "Buffer Addressing" (range checking) and §8.1.8 "Buffer Resource": out-of-range reads return zero and out-of-range writes are dropped, judged against the V#'s `num_records`.

### DIFF
--- a/src/graphics/host_gpu/renderer/renderDraw.cpp
+++ b/src/graphics/host_gpu/renderer/renderDraw.cpp
@@ -781,6 +781,7 @@ struct PreparedVertexBuffers {
 
 	std::array<vk::Buffer, MaxBuffers>     buffers {};
 	std::array<vk::DeviceSize, MaxBuffers> offsets {};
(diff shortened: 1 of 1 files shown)


brandostrong:
Closing: 3f80151a ("Respect vertex buffer sizes") makes the same change.

PR #467, opened 2026-09-02
loader: emulate the register forms of SSE4a EXTRQ and INSERTQ

brandostrong:
**Problem.** Astro Bot's asset unpacker executes register-form `extrq xmm2, xmm5` (`66 0f 79 d5`) and `insertq`. The emulator handled only the immediate forms, so the instruction trapped. A first attempt with swapped operands corrupted a register and crashed about a thousand flips later in libc.

**Change.** Emulate the register forms per the AMD64 SSE4a specification: destination is ModRM.reg, the length/index descriptor is ModRM.rm, and the destination's upper qword is left untouched as the hardware does.

**Effect.** Astro Bot's unpack stage runs through with no register corruption (six further runs).

**Verification.** Verified against the game's own unpacker disassembly; loader tests pass.

**References**
- AMD64 Architecture Programmer's Manual, Volume 4 (128-bit and 256-bit media instructions): `EXTRQ` (`66 0F 78 /0 ib ib`, `66 0F 79 /r`) and `INSERTQ` (`F2 0F 78 /r ib ib`, `F2 0F 79 /r`); destination is ModRM.reg, upper qword unchanged.

### DIFF
--- a/src/loader/x64InstructionEmulator.cpp
+++ b/src/loader/x64InstructionEmulator.cpp
@@ -515,7 +515,7 @@ static bool TryEmulateSse4a(PCONTEXT context) {
 		offset++;
 	}
 
-	if (rip[offset] != 0x0f || rip[offset + 1] != 0x78) {
+	if (rip[offset] != 0x0f || (rip[offset + 1] != 0x78 && rip[offset + 1] != 0x79)) {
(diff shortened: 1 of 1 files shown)


Nmzik:
Credited

PR #470, opened 2026-09-02
shader: model EXEC and VCC words as subgroup ballots

brandostrong:
**Problem.** EXEC and VCC words were modelled per lane, so branches on EXECZ/VCCZ and scalar arithmetic on the masks saw one lane's bit rather than the wave-wide word the hardware has. Wave32 shaders in Astro Bot took the wrong branches.

**Change.** Two commits. First: `BallotMask` emits a real subgroup ballot, `ThreadBit` tests `(word >> lane) & 1`, the entry block seeds EXEC with `Ballot(true)`, and in wave32 `vcc_hi`/`exec_hi` are ordinary registers. S

econd: EXECZ/VCCZ branches and the `execz`/`vccz` operands read the whole-wave value.

**Effect.** Astro Bot's wave32 shaders branch as the hardware does; needed to reach the game world.

**Verification.** Suites pass; the `wqm` SPIR-V budget test updated for the one extra entry ballot.

**References**
- AMD RDNA 2 Instruction Set Architecture Reference Guide, §3.3 "EXECute Mask", §3.9 "Vector Compares: VCC and VCCZ" and §4.2 "Branching" (`S_CBRANCH_EXECZ`/`VCCZ`): EXEC and VCC are wave-wide masks and the Z flags test the whole word.

### DIFF
--- a/src/graphics/shader/recompiler/frontend/translate/Translate.cpp
+++ b/src/graphics/shader/recompiler/frontend/translate/Translate.cpp
@@ -110,8 +110,8 @@ Decoder::Operand Translator::PlainOperand(const Decoder::Operand& operand) {
 }
 
 std::array<IR::U32, 2> Translator::BallotMask(IR::U1 value) {
-	return {ir.Select(value, IR::U32(IR::Value(1u)), IR::U32(IR::Value(0u))),
-	        IR::U32(IR::Value(0u))};
+	const auto ballot = ir.Emit(IR::ValueOpcode::Ballot, {value});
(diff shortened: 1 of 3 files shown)


brandostrong:
This seems to fail against `tests/shaderCfgTests.cpp`, the EXEC/SCC shared-arm case, though it is more a design disagreement than anything else. Up to you.

brandostrong:
Closing: 6cc9e5de ("shader: fix wave32 VCC halves and scalar mask moves") reimplements this.

PR #474, opened 2026-09-02
shader: lower SAMPLE_C to a sample and an ALU compare

brandostrong:
**Problem.** `IMAGE_SAMPLE_C*` and `IMAGE_GATHER4_C*` were emitted as Vulkan Dref ops with a comparison sampler. The host samples depth through colour-format views (R16/R32), which do not carry the depth-comparison feature, so every such draw or dispatch is invalid (`VUID-vkCmdDispatch-None-06479` in Astro Bot's volumetric-fog compute shader, and the same VUID on a draw).

**Change.** Sample with the plain opcode, take the red channel and compare it in ALU against the reference using the guest sampler's compare function (dword 0 bits 12–14: 0 passes the texel through, 7 is always-pass), selecting 1.0 or 0.0. Gathers compare each of the four texels. Comparison bits are stripped from every host sampler since no host compare remains. Test expectations updated from Dref ops to the plain sample and select.

**Effect.** Astro Bot's fog-volume writer runs hundreds of dispatches instead of stopping at its third; the validation error is gone on both the compute and the draw path.

**Credit.** The SAMPLE_C lowering follows the approach in #241 by @Pouare514; this is a re-implementation against the current emitter, kept to that one topic.

**Verification.** shader_cfg, resource_tracking, scalar_provenance pass; the compute suite reaches the same pre-existing stencil-discovery failure as main.

**References**
- AMD RDNA 2 Instruction Set Architecture Reference Guide, §8.2.7 "Image Sampler": S# dword 0 bits 14:12 `depth compare func` (0 never … 7 always).
- Vulkan specification: `VUID-vkCmdDispatch-None-06479` / `VUID-vkCmdDraw-None-06479`, Dref sampling requires `VK_FORMAT_FEATURE_2_SAMPLED_IMAGE_DEPTH_COMPARISON_BIT`, which colour formats lack.
- Approach from #241 by @Pouare514.

### DIFF
--- a/src/graphics/host_gpu/renderer/pipeline/descriptors.cpp
+++ b/src/graphics/host_gpu/renderer/pipeline/descriptors.cpp
@@ -806,15 +806,8 @@ static vk::Sampler NativeSampler(RenderContext&                       context,
                                  const ShaderRecompiler::IR::DescriptorValue& value) {
 	ShaderSamplerResource descriptor;
 	CopyNativeDescriptor(value, descriptor.fields);
-	const bool depth_compare = std::any_of(program.info.sampled_pairs.begin(),
-	                                       program.info.sampled_pairs.end(), [&](const auto& pair) {
-		                                       return pair.sampler == index &&
-		                                              pair.image < program.info.images.size() &&
-		                                              program.info.images[pair.image].depth_compare;
-	                                       });
-	if (!depth_compare) {
-		descriptor.fields[0] &= ~(0x7u << 12u);
-	}
+	// SAMPLE_C is lowered to a sample plus a compare in the shader; no host sampler compares.
+	descriptor.fields[0] &= ~(0x7u << 12u);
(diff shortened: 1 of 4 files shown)


Gih-pt:
@brandostrong An off-topic question: with all these PRs, does the game actually get to the menus?

brandostrong:
> @brandostrong An off-topic question: with all these PRs, does the game actually get to the menus?

Half yes, there is some raytracing related code that isn't in any of these recent PRs that is a blocker. I'm thinking of making a raytracing-only branch that follows main for those if you wish to pull from those commits. It'll be something I return to every so often when I'm stuck on normal rasterization.
Most of these PRs fix numerous crashes related to reaching the game world. The menu is reachable, so is save select, and so is the initial cutscene before actual gameplay after selecting your save slot, for me at least.

Nmzik:
Wrong implementation. RDNA2 defines _C as PCF—percentage-closer filtering

PR #478, opened 2026-09-04
Add update checker to launcher

DistantMyth:


DistantMyth:
- Checks the website update feed (kytyps5.github.io/data/updates.json) on startup with a fallback to GitHub releases.
- Shows a simple dialog with the changelog and download buttons for your OS when a new version drops.
- Added a "Check for updates" link right by the version label in the UI.
(btw, since the website itself refreshes it's data every 30 minutes, there might be a lag as long as half an hour before the user gets to know about the newer release, insignificant, but worth talking)

PR #490, opened 2026-09-05
shader: IMAGE_ATOMIC_FMIN/FMAX (MIMG 0x1e/0x1f)

francoisatt:
Ghost of Yotei (PPSA26344) fails shader recompilation on MIMG 0x1f — see
#281. Its first boot compute shader uses IMAGE_ATOMIC_FMIN/FMAX on an R32F
storage image.

### Credit

This is a port of #338 by @foufouadi, which was closed without being merged.
The analysis, the CAS-loop approach and the descriptor-format insight are
theirs. #338 no longer applies to main — the resource model was reworked since
(ResourceKind::StorageImageUint replaced by numeric_class /
TextureNumericClass, ValidateResourceSnapshot / MaterializeResources replaced
by MaterializeSnapshot / BuildResourceSpecialization), so a rebase produced
four structural conflicts. This PR redoes the same work against the current
model, and covers only the IMAGE_ATOMIC part — the DPP8 half of #338 is a
separate concern.

### Changes

decode — MIMG_ATOMIC_OPCODE_LIST gains 0x1e/0x1f, Opcode enum gains
IMAGE_ATOMIC_FMIN/FMAX (inserted after IMAGE_ATOMIC_XOR so the existing
SWAP..XOR range checks stay intact; the range bound in Memory.cpp is widened
to FMAX, which covers exactly the nine atomics).

translate/IR — ImageAtomicFMin32 / ImageAtomicFMax32, same signature as the
integer image atomics, classified as {ImageAccess::Atomic,
ImageResourceClass::Storage} so HasSideEffects reports correctly.

resources — atomic storage images may now carry k32Float. numeric_class stays
Uint for atomic images: ImageType asserts on it, IsSupportedStorageImageResource
requires it, and OpImageTexelPointer needs a u32 pointee to match the image's
sampled type. The float semantics live entirely in the bitcast inside the CAS
loop. The host view is forced to R32Uint for atomic resources — a legal
reinterpretation at equal width, matching the R32ui declared in SPIR-V.

SPIR-V — OpImageTexelPointer (u32) then AtomicUpdate + EmitFloatAtomicReplacement,
the same lowering the buffer and shared float atomics already use. AtomicUpdate
already handles ResourceKind::Image (device scope, ImageMemory semantics), so no
extra barrier and no SPV_EXT_shader_atomic_float dependency.

### Testing

Build: clean build on Windows, clang-cl 19.1.5, Ninja — no new warnings.
Tests: 36/36 pass, rebased onto 74a78f3.

**Not verified at runtime.** I don't have the title, so I cannot confirm the
shader compiles or renders correctly on real data. #338 reported 0 shaders
compiled before / 2 after for this change, but I cannot reproduce that. Runtime
validation from someone who can run the game would be welcome.

### AI assistance

Written with AI assistance. The analysis of the current resource model, the
numeric_class decision and the SPIR-V lowering were worked out in conversation
with Claude; every change was reviewed and built and tested locally by me.


francoisatt:
Rebased onto 6aa6e39, conflict in descriptors.cpp resolved — main had added
depth-format handling in the same expression, both changes kept.

Note: `kernel_file_system` fails on current main without any of my commits
(`guest PEEK and WAITALL preserve the wake bytes`, reproducible on a clean
checkout of 6aa6e39). Unrelated to this PR.

francoisatt:
Rebased onto c6f6e30. 35/36 tests pass — `kernel_file_system` still fails on
clean main (see #505), unrelated.

francoisatt:
This is conflicting now: main refactored the MIMG atomic table since I rebased — MimgAtomicInfo with its name strings became Detail::OpcodeMap with two elements. Only two files actually conflict (ImageOps.cpp and spirvEmitterImage.cpp), but the table change means the decode entries need rewriting against the new shape rather than just replaying the rebase.

Before I spend another evening on that: is this something you'd want? The PR has been open and CI-green for 8 days with no review, and this is already the second port of the same work — #338 died the same way. Ghost of Yotei (PPSA26344) is still blocked on MIMG 0x1f per #281.

Happy to redo it if there's interest. If the approach isn't wanted, say so and I'll close it — either answer is more useful to me than the PR sitting here.

bipinkrish:
@francoisatt similar thing is being implemented here https://github.com/KytyPS5/KytyPS5/pull/497 with much more improvements

PR #493, opened 2026-09-06
shader: IMAGE_BVH_INTERSECT_RAY behind --stub-bvh (MIMG 0xe6/0xe7)

francoisatt:
MIMG 0xe6 blocks five titles in #281 — Astro Bot (PPSA21564), Little Nightmares 2
(PPSA02154), Avatar: Frontiers of Pandora (PPSA01576), Severed Steel (PPSA05106),
and the original report. On Astro Bot the recompiler gets through 19 shaders
before hitting it.

Ray tracing is not implemented here. This adds decode plus an **inert
always-miss stub, behind an opt-in flag**. It removes one blocker so the next
one becomes visible; it does not make ray tracing work.

### The trade

Without `--stub-bvh` (the default) nothing changes: the shader still fails to
recompile, but with a message that says what to do —
`MIMG BVH opcode requires --stub-bvh (ray tracing is not implemented)`.

With the flag the shader compiles and the game runs past this point, but
rendering is wrong wherever the title uses ray tracing. A one-shot `LOGF`
warning is emitted the first time the stub fires so this isn't silent.

I've made it opt-in precisely because a silent stub that reports success is
worse than an honest failure. If you'd rather not have it at all, say so and
I'll drop the stub and keep only the decode — that alone turns the crash into a
clear message.

### Changes

**decode** — MIMG 0xe6 / 0xe7 added to `DecodeMimgOpcode`, two entries in the
`Opcode` enum placed outside the atomic range checks. Address-component counts
are 11/8 (0xe6) and 12/9 (0xe7) depending on A16, per the RDNA2 ISA;
`DecodeMimgAddressComponents` gains an `a16` parameter that other opcodes
return before ever reading.

**stub** — returns `{0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF}`. That
value is neutral under both readings of the result: as a child pointer on a box
node it's the "absent child" sentinel, so traversal terminates; as a float on a
triangle node it's a NaN, so every ordered comparison is false and no hit is
recorded regardless of how the shader reads `hit_status`. Zero would have been
wrong on both counts — child pointer 0 can be the root, and `t = 0.0` beats any
initial best.

**flag** — `--stub-bvh`, off by default. The gate sits at decode time so the
existing `SetUnsupported` path carries the message.

**tests** — `CheckBvhStub` covers both paths: with the flag off, the MIMG must
decode as `UNSUPPORTED` with a reason mentioning the flag; with it on, the full
pipeline must compile and the IR must contain exactly one
`ImageBvhIntersectRay`.

### Two things worth flagging in review

The decode gate reads `Config::BvhStubEnabled()` directly from the recompiler,
which is not how options normally reach it — `CompileOptions` is. I went this
way because `DecodeProgram(code, program)` is a free function with no options
struct, and threading one through would touch 22 call sites in the tests. The
`&&` short-circuit means Config is only read for a BVH opcode. Happy to do it
the other way if you prefer the recompiler stay free of Config.

In the test, a minimal `[MIMG + s_endpgm]` shader produces an empty IR: nothing
reads the destination VGPR, so `EliminateDeadCode` removes the stub along with
it. The test appends a buffer store to keep the result live. Real shaders
consume the result, so this doesn't affect them — noting it because the
first version of the test passed while proving nothing.

### Testing

Build: clean build on Windows, clang-cl 19.1.5, Ninja — no new warnings.
Tests: 36/36, on `74a78f3`.
Decode verified by hand against three real encodings from #281 (Astro Bot,
Little Nightmares 2, Severed Steel): opcode, dword count, address register
count and dmask all match.

**Not verified at runtime.** I don't have any of the affected titles, so I
can't confirm a game actually gets further with the flag on. Someone who can
run Astro Bot would settle that quickly — and would also reveal what the next
blocker is, which is the point of the flag.

### AI assistance

Written with AI assistance. The RDNA2 semantics, the choice of sentinel value
and the test design were worked out in conversation with Claude; every change
was reviewed, built and tested locally by me.


francoisatt:
Rebased onto 6aa6e39, no conflicts. 35/36 tests pass — the only failure is
`kernel_file_system`, which fails on clean main as well (see #505). Unrelated
to this PR.

Randomuser8219:
No change on Astro Bot with the new upstream.
```
--- Build ---
Source build 4b137e4
--- Stack Trace ---
[0] 0000000140ba3932
[1] 00000001407b359d
[2] 00000001407b2fb1
[3] 00000001407c746f
[4] 0000000140739c3e
[5] 0000000140710d97
[6] 0000000140737bed
[7] 00000001407370c4
[8] 000000014072f172
[9] 000000014073cf5c
[10] 00007ffce4abcd30
[11] 00007ffce6f0ccb7
[12] 00007ffce792caec
--- Fatal Error ---
Error: condition (!ShaderRecompiler::IR::MaterializeResources(resource_plan, runtime, resources, specialization)) is true in D:\a\KytyPS5\KytyPS5\src\graphics\host_gpu\renderer\pipeline\pipelineCache.cpp:360
```

FirasDev:
Tested against Astro Bot (ASTRO BOT V01.018.000, `PPSA21567`, EUR release, one of the five titles from #281). Note #281/#204 tested `PPSA21564`, which redump.org identifies as the NA release of the same v01.018 build, so same game/version, different regional title ID.

**Without `--stub-bvh`:** confirms the new message fires as intended: the old generic "MIMG opcode is not implemented" is now `MIMG BVH opcode requires --stub-bvh (ray tracing is not implemented)`, same PC/opcode as before.

**With `--stub-bvh`:** the shader that was blocking at PC `0x2190` (MIMG `0xe6`) compiles and the emulator gets past it, the decode + stub does what it says. But it doesn't reach the menu. It hits a second, unrelated crash a few shaders later:

```
Error: condition (!ShaderRecompiler::IR::MaterializeResources(resource_plan, runtime, resources, specialization)) is true
in pipelineCache.cpp:360
```

Symbolized call chain (Release build, `/DYNAMICBASE:NO`, so addresses resolved directly against the matching PDB):

```
PipelineCache::ProgramCache::Get<ShaderComputeInputInfo>   (pipelineCache.cpp:360, the EXIT_IF site)
PipelineCache::GetComputeProgram
RenderExecutor::DispatchDirect
CommandProcessor::DispatchDirect
CommandProcessor::DispatchIndirect
CpOpDispatchIndirect
CommandProcessor::ProcessPm4 / Process
GuestGpu::Process / ThreadRun
```

So this is a compute shader, reached via an indirect dispatch (`CpOpDispatchIndirect`), that also uses `IMAGE_BVH_INTERSECT_RAY`.

Edit: my original theory here was wrong, leaving this in place with the correction rather than pretending I got it right the first time. I'd guessed resource-plan extraction was treating the BVH instruction's resource operand as a normal image descriptor and rejecting it. Checked the actual stub code (Memory.cpp, IMAGE_BVH_INTERSECT_RAY / IMAGE_BVH64_INTERSECT_RAY case):

```cpp
const auto memory = MemoryInfoFromDecoded(inst);
const auto result = ir.Emit(IR::ValueOpcode::ImageBvhIntersectRay);
WriteImageComponents(inst.dst, result, memory, 4u);
```

`ir.Emit(IR::ValueOpcode::ImageBvhIntersectRay)` takes no arguments. Unlike `IMAGE_SAMPLE`/`IMAGE_LOAD`'s handlers, it never touches `inst.src1`/`inst.src2` (SRSRC/SSAMP), so it never registers a resource anywhere. I also checked `ExtractResourcePlan` directly: it doesn't classify opcodes itself, it just copies whatever's already in `program.info.buffers/images/samplers`, which gets populated during translation by the resource-registering opcodes. Since the stub never calls that registration path, there's no BVH-shaped descriptor for `MaterializeResources` to see, let alone misclassify.

So the second crash is caused by something else in that shader, unrelated to the BVH instruction's resource operand. Planning to rerun with `--shader-log-direction File` to get the actual failing shader's IR instead of guessing again.

Net result: at least for this title, `--stub-bvh` removes the blocker it targets but doesn't get the game running. There's a second, separate crash in resource materialization for this shader, not yet confirmed whether it's actually related to the BVH instruction or just another pre-existing gap that happened to be next in line. No idea whether the other four titles in #281 hit the same wall or get further.

Tested on: Windows, clang-cl 19.1.5, Release, branch built at `adc8e95`.

francoisatt:
Thanks both — this is exactly the validation I couldn't do myself.

@FirasDev your correction is right, and it matches what I found earlier:
ImageBvhIntersectRay isn't classified in ImageOpcodeInfoOf (ValueOpcodes.cpp),
so ResourceTracking::Collect returns early and never calls AddImage. The stub
registers no resource at all — nothing for MaterializeResources to reject.

Worth noting for whoever picks up that second crash: that same gap means the
BVH descriptor never reaches the host as a resolved resource. It'll matter for
a real implementation, which needs the descriptor's base address to read nodes.
Classifying the opcode would be part of that work, not of this PR.

The --shader-log-direction File run sounds like the right next step. Happy to
look at the IR if you post it.

francoisatt:
@FirasDev main has moved a fair bit since adc8e95 — #457 (dense image resource
limit raised to 64) landed in c6f6e30 and touches image resources, which is
where your second crash lands. Might be worth a retest on that build before
digging into the IR. I've rebased this branch onto c6f6e30.

FirasDev:
@francoisatt retested on the rebase (3b16ede onto c6f6e30). Same crash, same file:line, and the stack frames from CommandProcessor down to thread entry are byte-identical to the pre-rebase trace posted above (only frames 0-4 shifted, consistent with just recompilation moving code around inside pipelineCache.cpp itself).

```
--- Build ---
Source build 3b16ede
--- Stack Trace ---
[0] 0000000140488099
[1] 0000000140487fc2
[2] 00000001400a78d8
[3] 00000001400a733c
[4] 00000001400bc4ec
[5] 000000014002ceb7
[6] 000000014002cfda
[7] 000000014000b5b5
[8] 000000014002b1bc
[9] 000000014002a580
[10] 000000014002a048
[11] 0000000140026678
[12] 000000014002f8fc
--- Fatal Error ---
Error: condition (!ShaderRecompiler::IR::MaterializeResources(resource_plan, runtime, resources, specialization)) is true in pipelineCache.cpp:360
```

francoisatt:
Thanks for the retest — that rules out #457 cleanly.

MaterializeResources fails through SpecializationFail, which formats an actual
reason string ("... uses unsupported format {}", "... unsupported resource
class", etc). That message is almost certainly being swallowed: printf_direction
defaults to Silent, so every LOGF is dropped.

Worth a run with `--printf-direction Console` before anything else — the line
just before the EXIT should name which descriptor and why. That turns this from
guesswork into a one-line answer, and it's cheaper than dumping the IR.

If it does need the IR after that, --shader-log-direction File is the right
next step.

Randomuser8219:
> Thanks for the retest — that rules out #457 cleanly.
> 
> MaterializeResources fails through SpecializationFail, which formats an actual reason string ("... uses unsupported format {}", "... unsupported resource class", etc). That message is almost certainly being swallowed: printf_direction defaults to Silent, so every LOGF is dropped.
> 
> Worth a run with `--printf-direction Console` before anything else — the line just before the EXIT should name which descriptor and why. That turns this from guesswork into a one-line answer, and it's cheaper than dumping the IR.
> 
> If it does need the IR after that, --shader-log-direction File is the right next step.

image
I tried with Astro Bot, nothing particularly interesting with printf set to "Console".

francoisatt:
Superseded by #558, which fixes the actual root cause (the SRT plan bug) and
implements real traversal rather than a stub. Closing — the stub was only ever
meant to make the next blocker visible, and it did its job.

Issue #505, opened 2026-09-07
kernel_file_system test fails on main: guest PEEK and WAITALL preserve the wake bytes


francoisatt:
On a clean checkout of main (6aa6e39), Windows x64, clang-cl 19.1.5, Ninja:

    16/36 Test #16: kernel_file_system ... Exit code 0xc0000409
    KernelFileSystemTests: failed: guest PEEK and WAITALL preserve the wake bytes

Reproducible on every run, no local changes. The other 35 tests pass.

Found while rebasing #490 — unrelated to that PR, it reproduces on main alone.

NaN-Error:
Hi! I used ChatGPT's Astra Pro to investigate this issue and wanted to share a possible cause for review.

The AI-assisted analysis points to the receive path in `src/libs/network.cpp` at the reported commit, `6aa6e39`. The guest's `PEEK | WAITALL` flags appear to be translated into `MSG_PEEK | MSG_WAITALL` and passed through to Winsock. Microsoft's `recv` documentation states that this combination fails with `WSAEOPNOTSUPP`, which seems consistent with the Windows failure reported here.

Reference:
https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-recv

Checking `WSAGetLastError()` immediately after the failing native receive would help confirm whether this is the cause.

One possible approach to evaluate would be a Windows-only adjustment in `Libs::Network::Net::Recvfrom` that preserves `MSG_PEEK` but clears `MSG_WAITALL` when both are present. This would leave ordinary `WAITALL`-only receives unchanged and avoid modifying the shared flag-conversion function. However, someone familiar with the expected guest socket behavior would need to evaluate whether that is appropriate, particularly for partial reads and blocking behavior.

To be transparent, this investigation was AI-assisted and did not include native Windows execution or the complete emulator test suite. Please treat this as a possible explanation and proposed direction, not a confirmed or regression-tested fix.

Would this kind of AI-assisted investigation be helpful to the project? I'd be interested in continuing to examine selected issues with AI and sharing focused findings, supporting references, and possible solutions for a maintainer or contributor to evaluate and implement when appropriate.

My goal would be to help narrow down problems, not assume that AI-generated conclusions are correct or shift a large amount of verification work onto the maintainers. Before continuing, I'd appreciate your guidance on whether you welcome this kind of contribution and what level of validation or format you would expect.

Thanks for your work on the project!

PR #511, opened 2026-09-08
loader: refresh imports after dynamic module loads

### LINKED ISSUE #307: [GAME STATUS]: Castlevania Dominus Collection crashes after game selection
### Game title

Castlevania Dominus Collection

### Game ID / serial

PPSA15645

### KytyPS5 version

Source build b9480fa-dirty (Release)

### Compatibility status

Main menu

No known working build

### Result details

The collection boots successfully and reaches its game-selection menu at approximately 60 FPS. Menu rendering, audio, and keyboard input work.

Selecting the highlighted game closes Kyty immediately. Windows Application Error records a repeatable execute access violation:

- Exception code: `0xc0000005`
- Fault address/offset: `0x0000000000000000`
- Faulting module: unknown
- Reproduced twice with the same failure

The strict loader log ends during the second dynamic load/relocation of `dra03.prx`. The process terminates before Kyty can flush a crash record to the log. Local minidump inspection confirms `RIP=0`.

### Steps to reproduce the result

1. Open KytyPS5 Launcher.
2. Boot Castlevania Dominus Collection (PPSA15645, v01.003).
3. Wait for the collection game-selection menu.
4. Press J to select the highlighted game.
5. Kyty closes immediately while loading the selected game.

### Expected behavior

The selected Castlevania game should finish loading and become playable.

### OS

Windows 11 Pro 25H2, build 26220.9223

### CPU

AMD Ryzen 9 7900X3D (24 threads)

### GPU

NVIDIA GeForce RTX 4080

### RAM / VRAM

32 GB RAM / 16 GB VRAM

### Log file upload

Full sanitized log archive prepared as castlevania-dominus-crash.zip; it will be attached by the reporter immediately after issue creation.

### Screenshots/video

Image

Image

### Extra notes

Game version: 01.003.000
Content ID: JP0101-PPSA15645_00-DOMINUSCOLLECTIO
Default launcher settings. No PSN or official network service was used.
Two Windows crash reports and two local minidumps were produced. The attached ZIP contains the sanitized complete strict log and a crash summary.

OldKrab:
Castlevania Dominus Collection crashes after unloading and reloading `dra03.prx` because the main executable retains an invalid address for an imported object.

`UnloadProgram()` already calls `RelocateAll()`. Once the module is removed, this replaces its object's address in the main executable's GOT with an invalid-object placeholder. Loading the module again currently relocates only the newly loaded module, leaving that GOT entry unchanged.

I propose calling `RelocateAll()` from `RelocateProgram()` as well, so existing imports receive the loaded module's addresses before it starts.

Validation: Linux build passed. Order of Ecclesia passed the original crash point and ran for roughly five minutes without another crash. Windows and audio remain untested.

Fixes #307.

### DIFF
--- a/src/loader/runtimeLinker.cpp
+++ b/src/loader/runtimeLinker.cpp
@@ -1288,7 +1288,7 @@ void RuntimeLinker::RelocateProgram(Program* program) {
 	EXIT_IF(program == nullptr);
 	EXIT_IF(std::find(m_programs.begin(), m_programs.end(), program) == m_programs.end());
 
-	Relocate(program);
+	RelocateAll();
 	if (!GamePatch::ApplyPending(program)) {
 		EXIT("Failed to apply pending game cheat\n");
 	}


francoisatt:
Built and tested on Windows (clang-cl 19.1.5, Ninja, clean build) at 0e2d667.
35/36 tests pass — the only failure is `kernel_file_system`, which fails on
clean main too (#505). No regression from this change.

I can't run Castlevania, so I can't confirm the fix itself — only that it
builds and doesn't break the suite on the platform you couldn't test.

Two things a reviewer will probably ask about, worth addressing up front:

- Cost: RelocateAll() re-runs over every program on each dynamic module load.
  GTA V has ~171k relocations in the main executable alone, so a title that
  loads modules frequently might notice.
- Overwriting: #343 (closed) tackled the same problem and explicitly avoided
  RelocateAll on the grounds that it "would rewrite every live import slot" —
  it added a classifier so only linker-owned pending values were touched. Your
  approach is much simpler; the question is whether any import slot can
  legitimately be modified by guest code and get clobbered here.

Also worth noting: SharpEmu does a dedicated pass for this and logs it as
`Imported data rebind: rebound=11, unresolved=0`, so the general idea is
sound — it's the scope that differs between the two approaches.

PR #532, opened 2026-09-09
renderer: finalize bindings after every range-changing operation

CheesyPoofs346:
**Problem.** A binding captured during draw preparation can be left pointing at a buffer that a later operation in the same draw retired. `NativeStorageBuffer` captures a `VkBuffer` handle, an offset and a packed byte adjustment; any later `ObtainBuffer` can reach `CreateBuffer`, which replaces the buffers covering a range with one wider buffer and retires the originals. Retirement is deferred past GPU completion, so the captured handle stays usable but stops being current: a later upload lands in the replacement, and a write through the stale alias lands in storage nothing else reads. `RebindBuffers(pixel)` can retire what `RebindBuffers(vertex)` captured, and `RebindImages` reaches `CreateBuffer` through `ResolveTexture -> FindImage -> InitializeImage -> ObtainBufferForImage`, so it can retire buffers both stages captured. The `shader_data` upload carries the packed memory offsets and ran before any of the image work, so the offsets and the handles could disagree.

**Change.** Split obtaining from publishing. Obtaining keeps its position and all of its side effects — synchronize, upload, dirty-state change, LRU touch — unchanged and unrepeated, and records how each binding was served. `PublishBuffers` then re-resolves every cache-backed binding from its **guest range** (a `BufferId` is exactly what retirement invalidates) through a new non-allocating `BufferCache::FindPublishedOwner`, and rewrites the descriptor, the packed offset and the `shader_data`/`flattened_srt` uploads from one settled snapshot. `FinalizeBindings` runs three passes over the whole stage span — rebind buffers, rebind images, publish — so no stage is finalized against a state a later stage changes; both the graphics and compute paths call it, so rebinding without publishing is not expressible. Stream-backed bindings are republished as issued: `ObtainBuffer`'s stream fast path returns a ring allocation holding a private copy that no guest range owns, so re-resolving one by address would discard it. A `published` flag is cleared by anything that can retire a buffer and asserted in `CommitBindings`.

**Effect.** No measured performance change; performance was not measured for this and no improvement is claimed. Publication adds a page-table lookup per cache-backed binding and moves two uploads later in the same draw. This closes a silent-corruption path: bindings that read bytes that were never uploaded to the buffer they point at, and writes that land in a retired buffer.

**Verification.** Builds clean on this base. The equivalent change in a downstream tree is covered by a device-backed regression that proves the defect by GPU readback — the captured handle reads the pre-update byte while the live owner reads the updated one, with a control check first so the stale result cannot be attributed to missing upload tracking — and additionally checks that a stream-backed binding keeps its allocation, offset and contents, that publication fails loudly rather than allocating when a range has no owner, and that rebinding clears publication. That tree also ran two full GTA V city drives with the change, exit code 0, no errors and no publication failures; the driver reported no visual problems.

**Limitations.** Marked draft for these:
- The regression above has **not** been ported to this base. Bringing the device harness up here crashes inside `ObtainBuffer` during test setup, unresolved, so this PR carries no automated regression. The fix itself is a re-implementation against this tree's structure, not the downstream patch.
- Not exercised anywhere: an image-triggered replacement specifically (retirement is driven by a spanning `ObtainBuffer` in the downstream test), writable aliases (`is_written` bindings), and `CommitBindings` itself — the `published` assertion is verified by reading, not by running.
- Reviewers may prefer the headless device setup used by `ShaderRecompilerComputeTests` over the SDL-window approach the downstream harness uses, if a regression is added here.

**References**

No external specification. Correctness argument only: `NativeStorageBuffer` captures `buffer->Handle()` into `prepared.buffers`, and `CreateBuffer` retires the buffers it replaces while `BindGraphicsResources` continues to rebind further stages and images afterwards.

### DIFF
--- a/src/graphics/host_gpu/renderer/cache/bufferCache.cpp
+++ b/src/graphics/host_gpu/renderer/cache/bufferCache.cpp
@@ -496,6 +496,21 @@ std::pair<Buffer*, uint64_t> BufferCache::ObtainBuffer(uint64_t vaddr, uint64_t 
 	return {buffer, buffer->Offset(vaddr)};
 }
 
+std::pair<Buffer*, uint64_t> BufferCache::FindPublishedOwner(uint64_t vaddr, uint64_t size) {
+	if (!GuestRange {vaddr, size}.Valid()) {
+		return {nullptr, 0};
+	}
+	const auto* owner = m_page_table.Find(vaddr >> PageTable::kPageBits);
+	if (owner == nullptr || !*owner) {
+		return {nullptr, 0};
+	}
+	auto& buffer = m_slot_buffers[*owner];
+	if (buffer.is_deleted || !buffer.IsInBounds(vaddr, size)) {
+		return {nullptr, 0};
+	}
+	return {&buffer, buffer.Offset(vaddr)};
+}
+
 std::pair<Buffer*, uint64_t> BufferCache::ObtainBufferForImage(uint64_t vaddr, uint64_t size) {
 	if (!GuestRange {vaddr, size}.Valid()) {
 		EXIT("BufferCache: invalid image source\n");
--- a/src/graphics/host_gpu/renderer/cache/bufferCache.h
+++ b/src/graphics/host_gpu/renderer/cache/bufferCache.h
@@ -55,6 +55,20 @@ public:
 		}
 		EXIT("BufferCache: invalid utility-buffer usage\n");
 	}
+	// Resolves the buffer that currently owns a guest range, or {nullptr, 0} when none does.
+	// Performs a page-table lookup and nothing else: no allocation, no join, no synchronize, no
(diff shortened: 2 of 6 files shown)


CheesyPoofs346:
Do not merge — this introduces visible geometry corruption.

A/B on the same scene and save, two builds differing only by this change: with it, large parts of interior geometry explode into flat dark shards (ceilings, walls, whole surfaces) while characters and most props render correctly; without it the same scene is clean. The corruption is selective, which points at one of the non-uniform paths in publication — the per-draw dedup hit path or formatted/texel buffers — rather than the offset arithmetic being wrong for every binding.

That also corrects the Effect section above: I wrote that performance was not measured, which was true, but I did not state that behaviour was unverified in a real scene. It is now verified to be wrong.

Investigating; will update or close.

PR #542, opened 2026-09-09
Fix/shader loader fixes

shadowbeat070:
4 miscellaneous to fix graphical issues and missing implemenations.

## Summary

- a6342b7 graphics: clear render target with over 32 layers:  The per layer metedata cleare mask was 32 bits wide which made a 64 wide slice skip the DCC clear entirely.

- 352bf3f shader: decode scalar conditional 32-bit moves: Adds the missing S_CMOV_B32 scalar opcode to the GCN decoder, so shaders using it no longer hit an unsupported instruction.

- 234d015 loader: see red zone kept in a frame-pointer frame: Red zone detection only recognized RSP as a stack pointer, so a leaf function keeping its red zone at [rbp - n] got no protection at all.

- c772ca2 shader: apply image_sample texel offsets: Fixes texel offsets being silently dropped on image_sample_*_o. SPIR-V cannot express a runtime offset on a sample: ConstOffset requires a compile-time constant, and the dynamic Offset operand is legal only for gather. 


Nmzik:
@shadowbeat070 All good. Please include list of games that these fixes apply :)

shadowbeat070:
@nmzik list of games include: the pathless, stray and silent hill the short message.

PR #543, opened 2026-09-09
graphics: support deferred DCC clears beyond 32 slices

MehmetCambaz:
## Problem

`MetaDataInfo::clear_mask` tracked pending DCC clear slices in a `uint32_t`, so only
slices 0-31 could be represented. `PrepareDccClear` bailed out on any view whose slice
range fell outside that window, and `IsMetaCleared`/`TouchMeta` reported slice 32 and
above as not pending. A recognised guest DCC clear on a layered target with more than
32 slices was armed and then never materialised, leaving the image with its previous
contents.

## Change

Track pending slices with the existing `RangeSet` across the slice range the guest can
address. Widening the integer would only move the limit — `CB_COLOR_VIEW` encodes 13-bit
slice indices.

`PrepareDccClear` now materialises the intersection of the bound view with the slices
still pending, one contiguous run at a time. Partial consumption is unchanged: slices
already cleared are not cleared again, since they may have been rendered into since, and
a repeated metadata fill re-arms the whole range. The metadata entry is re-found after
each `ClearImage`, because retiring an image can erase it.

HTile still arms the same 32 slices as before; widening that belongs with a separate
multi-layer depth change.

## Validation

New `texture_cache_dcc_slice_range` covers a 64-slice target, a view starting above slice
31, overlapping partial consumption, and a target past 64 slices so widening the mask
alone does not pass. The cases run through `ResolveRenderColorTarget`/`FindRenderTarget`
and check host texels and pending state before and after.

Focused DCC, HTile, depth and cache groups pass. Full CTest 37/38; `kernel_file_system`
fails identically on pristine upstream.

Kena: Bridge of Spirits exposed this with a 64-slice RGBA16F volumetric lighting target
that is fast-cleared each frame before additive light injection. Before, the clear never
landed and radiance accumulated across frames into progressively over-bright blue/white
lighting. After, lighting stays stable.

## AI assistance

AI assistance was used during investigation, implementation and testing. I reviewed the
resulting change and validation before submitting it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -577,6 +577,8 @@ if(BUILD_TESTING)
 		COMMAND $<TARGET_FILE:shader_recompiler_compute_tests> --depth-readback-only)
 	add_test(NAME compute_meta_clear_classification
 		COMMAND $<TARGET_FILE:shader_recompiler_compute_tests> --compute-meta-clear-only)
+	add_test(NAME texture_cache_dcc_slice_range
+		COMMAND $<TARGET_FILE:shader_recompiler_compute_tests> --dcc-slice-range-only)
 	add_test(NAME buffer_cache_dirty_gc
 		COMMAND $<TARGET_FILE:shader_recompiler_compute_tests> --buffer-cache-gc-only)
 
--- a/src/graphics/host_gpu/renderer/cache/textureCache.cpp
+++ b/src/graphics/host_gpu/renderer/cache/textureCache.cpp
@@ -5,6 +5,7 @@
 #include "common/logging/log.h"
 #include "common/profiler.h"
 #include "graphics/guest_gpu/gpu_format.h"
+#include "graphics/guest_gpu/pm4.h"
 #include "graphics/guest_gpu/tile.h"
 #include "graphics/host_gpu/graphicContext.h"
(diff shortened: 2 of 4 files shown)


Nmzik:
@MehmetCambaz Could you please retest with main branch if the issue still persists?

MehmetCambaz:
@Nmzik Retested on current main (e6156bef). I couldn’t reproduce the original DCC/lighting issue in the run, although the low FPS and unrelated late-loading issues made the visual check a bit difficult.

I also checked the current DCC path and this PR appears to have been superseded by 437e69ef (Remove extra DCC tracking). TrackDccFill, PrepareDccClear, and the pending DCC tracking this PR changes are gone, and MaterializeDccClear now works per slice against the actual layer count rather than the old 32-bit pending mask.

So this PR is obsolete against current main and I’ll close it. Thanks for pointing me back to the latest code.

Also, while testing Kena I’ve been working on a couple of separate performance improvements: #562 reduces the PrepareBda overhead substantially in Kena, and #628 reduces SRT/resource-materialization CPU cost. If you get a chance, I’d appreciate a look at those as well.

MehmetCambaz:
closing the pr

Issue #290, opened 2026-08-18
[GAME STATUS]: DOOM + DOOM II


m7med80742:
### Game title

DOOM + DOOM II

### Game ID / serial

PPSA21444

### KytyPS5 version

0.2.2 (KytyPS5-2026-08-18-6bf6929)

### Compatibility status

In game

### Result details

game is playable, menu text is broken

<img width="1282" height="752" alt="Image" src="https://github.com/user-attachments/assets/d6af69c9-ca7d-4aba-9a1f-d8f788d45268" />

<img width="1282" height="752" alt="Image" src="https://github.com/user-attachments/assets/816c8864-c52c-47f2-8ca7-73a321a46eae" />

### Steps to reproduce the result

1. Open the emulator
2. Boot the game
3. Play the game

### Expected behavior

the menu should load with correct text

### OS

Windows 11 pro 25H2

### CPU

AMD Ryzen 5 8600G

### GPU

AMD Radeon 780m, driver 26.7.1

### RAM / VRAM

16GB DDR5 RAM / Shared VRAM

### Log file upload

[_kyty DOOM 1&2.zip](https://github.com/user-attachments/files/31195357/_kyty.DOOM.1.2.zip)

### Screenshots/video

<img width="1282" height="752" alt="Image" src="https://github.com/user-attachments/assets/27b0a8cd-df28-422b-b5c6-602183c8a8f2" />

<img width="1282" height="752" alt="Image" src="https://github.com/user-attachments/assets/6d8306bc-7a29-4c4d-92cf-9636006efdfb" />

<img width="1282" height="752" alt="Image" src="https://github.com/user-attachments/assets/afd13f3f-6519-4587-9a8c-905a5afa595b" />

<img width="1282" height="752" alt="Image" src="https://github.com/user-attachments/assets/41c7e6b4-bf52-441b-b056-02cd7271fe38" />

PR #545, opened 2026-09-09
Fix stencil synchronization and missing DOOM menu text

OldKrab:
Related issue: [#290](https://github.com/KytyPS5/KytyPS5/issues/290).

DOOM + DOOM II displays white backgrounds around glyphs and loses menu letters. Stencil buffer contents were not uploaded into Vulkan images, and garbage collection could discard rendered glyph atlases when older buffer writes blocked readback.

This change:

- Uploads stencil contents and refreshes them after guest writes.
- Prevents pending uploads from undoing optimized stencil clears.
- Preserves stencil contents when images are replaced or garbage-collected.
- Retains potentially current atlas images while allowing GC to reach other reclaimable images.

Changed depth/stencil pairings remain unsupported: log once and continue, accepting possible rendering artifacts. Retained images may occupy GPU memory longer.

Validated on Linux/NVIDIA with DOOM startup and collection menus, the existing GPU cache suite, and new regressions for clear/rebind, stencil retirement, and GC progress. Each regression failed before its fix and passes afterward. Windows/AMD and long gameplay remain unverified.

### Screenshots

DOOM + DOOM II startup screen on Linux/NVIDIA, without debugger overrides.

| Before | After |
| --- | --- |
| Before: white glyph backgrounds and missing options | After: readable text and restored options |

### DIFF
--- a/src/graphics/host_gpu/renderer/cache/textureCache.cpp
+++ b/src/graphics/host_gpu/renderer/cache/textureCache.cpp
@@ -955,7 +955,7 @@ struct TextureCache::DownloadPlan {
 
 TextureCache::TextureTransferPlan
 TextureCache::BuildTextureTransfer(const Image& image, BindingType binding,
-                                    TransferDirection direction) const {
+                                   TransferDirection direction) const {
 	const auto& info             = image.info;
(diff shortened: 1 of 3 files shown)


Nmzik:
Credited

PR #551, opened 2026-09-09
graphics/window: fix device rejection on GPUs lacking depthBounds/barycentric, and input-driven stutter

Hultwl:
## Summary

Two independent fixes, tested together on Intel UHD 620 (KBL, Mesa anv, Arch Linux):

1. **Device rejection on GPUs without `depthBounds`/`VK_KHR_fragment_shader_barycentric`.** Both were hard requirements in `VulkanFindPhysicalDevice`, causing `Could not find suitable device` on hardware that's otherwise fully capable. This commit adds fallback approaches for both.

2. **Input-driven stutter.** The main loop's idle wait was an unbounded `SDL_WaitEvent()`, so cross-thread work queued via `RunOnMainThread()` only got drained when an SDL event happened to arrive. Bounded to a 4ms timeout.

## Testing

- Before: immediate `EXIT("Could not find suitable device")` on boot, confirmed via a full `VkPhysicalDeviceFeatures2`/extension dump against the actual suitability-check requirements.
- After: boots, compiles shaders, and runs **Dreaming Sarah** to completion with no crashes.
- Stutter fix confirmed by observation (stutter present before, gone after, independent of mouse movement).

## Caveats worth flagging for review

- The `fragmentShaderBarycentric` fallback is an **approximation**: it treats every `GetInterpolationParameter` read as reading the hardware-interpolated value rather than reconstructing GCN's explicit per-vertex delta math. This matches real-world shader usage (compilers emit these instructions to reconstruct standard interpolation, not exotic per-primitive effects) and is the same approach used by SharpEmu's `Gen5SpirvTranslator.TryEmitInterpolation`, but I don't have a way to verify bit-exact correctness against real hardware for edge cases that might genuinely rely on the delta math. Flagging this explicitly rather than presenting it as a guaranteed-correct implementation.
- Only tested against one GPU (Intel UHD 620) and one game so far.

---

**Update:** Added a follow-up commit that scopes the fragmentShaderBarycentric
fallback to only affect the GetInterpolationParameter (v_interp) approximation
path, without hard-blocking pixel shaders that use PS's perspective-center
VGPR feature (a separate hardware mechanism unrelated to v_interp). The
latter has no equivalent approximation, but empirically works fine on Mesa
ANV even without the extension formally enabled, so it's left permissive
rather than refusing to run. Tested working again on Intel UHD 620 with
Dreaming Sarah.

### DIFF
--- a/src/graphics/host_gpu/graphicContext.h
+++ b/src/graphics/host_gpu/graphicContext.h
@@ -30,6 +30,9 @@ struct GraphicContext {
 	bool                               rt_extensions_enabled                 = false;
 	bool                               compute_subgroup_size_control_enabled = false;
 	bool                               sample_rate_shading_enabled           = false;
+	bool                               depth_bounds_supported                = false;
+	// True when VK_KHR_fragment_shader_barycentric is available.
+	bool                               barycentric_supported                 = false;
 	bool                               attachment_feedback_loop_enabled      = false;
 	bool                               provoking_vertex_last_enabled         = false;
 	bool                               supports_block_texel_view              = false;
--- a/src/graphics/host_gpu/renderer/pipeline/pipelineCache.cpp
+++ b/src/graphics/host_gpu/renderer/pipeline/pipelineCache.cpp
@@ -342,6 +342,7 @@ struct PipelineCache::ProgramCache {
 		options.early_dump  = options.dump_ir;
 		options.dump_label  = label;
 		options.input_info  = stage_input;
+		options.barycentric_supported = barycentric_supported;
 		options.scratch_dwords = input_info.scratch_size_dwords;
 		if constexpr (std::is_same_v<InputInfo, ShaderVertexInputInfo>) {
 			options.user_data_base = 8;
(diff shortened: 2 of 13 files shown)


Hultwl:
The fallback in the previous commit only covers GetInterpolationParameter
(GCN v_interp instructions) reads. There's a second, separate source of
barycentric requirement — ps_perspective_center_vgpr, a PS input config
where the hardware delivers interpolation weights directly into VGPRs at
wave entry, independent of any v_interp instructions in the shader body.
Games using it (e.g. Dreaming Sarah) hit a hard EXIT_NOT_IMPLEMENTED on
devices without the extension, even though the shader ran fine before the
guard was actually wired up to a real value (previously it was dead code).

There's no equivalent approximation for that path the way there is for
v_interp, but Mesa ANV tolerates the SPIR-V capability declaration even
without VK_KHR_fragment_shader_barycentric formally enabled on the device,
so this leaves the emitter guard permissive rather than blocking. Confirmed
working again on Intel UHD 620.

PR #555, opened 2026-09-10
shader: fix BVH resource classification and specialization crash

FirasDev:
This builds on #493 by @frans (cherry-picked as-is: the opcode decode, the stub, the --stub-bvh flag, and his tests), plus two fixes I found by running Astro Bot (PPSA21567) against it with --stub-bvh enabled.

Fix 1: the BVH resource was classified wrong. The stub registered its resource operand as a buffer resource, which routes it through SRT specialization to resolve the descriptor as a compile-time constant. But the BVH node-pool base address is a runtime pointer (two SGPRs read directly), not something specialization can resolve, so it crashed immediately. This reads it the same way S_LOAD's raw-address path already does, and leaves it unregistered with resource tracking, so nothing tries to constant-fold it.

Fix 2: resource specialization aborted the whole process on a control-flow-dependent descriptor. With fix 1 in, Astro Bot got much further, then hit an unrelated crash: some other shader (not BVH) computes part of a descriptor differently depending on runtime control flow, and ResolveInvariantPhi correctly detects that divergence, but the caller treated it as fatal for the entire shader. That's the same kind of "descriptor doesn't make sense" situation NullImageDescriptor/ValidImageDescriptor already tolerate elsewhere in this file, so this applies the same tolerance: zero out just that one resource and keep going instead of aborting.

Testing: ran against Astro Bot (PPSA21567) with --stub-bvh. Before these fixes: hard crash before the BVH instruction is even reached. After: compiles far more shaders than before, then hits a separate, unrelated access violation in the PS5 module loader (runtimeLinker.cpp) that has nothing to do with either fix or with ray tracing. Astro Bot does not fully boot yet, these two fixes remove two real crashes along the way.

This also carries the from-scratch BVH parsing/walking work (src/graphics/rt/), a standalone test confirming the host GPU can create an RT-capable Vulkan device, and a working bottom-level acceleration structure builder verified to actually build a BLAS on real hardware. None of that is wired into the shader pipeline yet, still blocked on real ray-query codegen, it's groundwork for that.

AI disclosure: I used Claude to help investigate, implement, and test this. I reviewed and ran everything against real hardware and a real game before submitting.


rodjunger:
Some comments are in french. Please at least read what Claude has generated before submitting a PR.

src/graphics/shader/recompiler/frontend/translate/Memory.cpp
```diff
@@ -1049,6 +1052,27 @@ bool Translator::EmitMemory(const Decoder::Instruction& inst) {
 		case Decoder::Opcode::IMAGE_GATHER4_C_LZ_O:
 		case Decoder::Opcode::IMAGE_GATHER4H: return IMAGE_GATHER(inst);
 
+		case Decoder::Opcode::IMAGE_BVH_INTERSECT_RAY:
+		case Decoder::Opcode::IMAGE_BVH64_INTERSECT_RAY: {
+			// Stub : sémantique inerte « aucune intersection » — 4 canaux 0xFFFFFFFF
+			// (pointeur d'enfant absent pour un nœud boîte / temps NaN pour un nœud triangle)
```
rodjunger: french comment

FirasDev:
> Some comments are in french. Please at least read what Claude has generated before submitting a PR.

Those are unmodified, from @francoisatt's own commits in #493 (c1befc761d, 3b16edede2), cherry-picked as-is like the PR description says, would you like me to translate them ?

rodjunger:
> > Some comments are in french. Please at least read what Claude has generated before submitting a PR.
> 
> Those are unmodified, from @francoisatt's own commits in #493 ([c1befc7](https://github.com/KytyPS5/KytyPS5/commit/c1befc761d3c363c0de596e9cdb05100a1b41f90), [3b16ede](https://github.com/KytyPS5/KytyPS5/commit/3b16edede2ce5173e9f25f906692a17644fed65c)), cherry-picked as-is like the PR description says, would you like me to translate them ?

I'm not an official maintainer but I don't see how a maintainer or anyone would want comments in french, so it's good to update them to english. The next thing that looks off to me is the formatting, I don't know if it's a github bug but most files seem to have really broken formatting, running a linter/formatter wouldn't hurt

francoisatt:
@FirasDev sorry about the French comments — those were mine, thanks for translating them. I'll keep code comments in English from now on.

PR #557, opened 2026-09-10
Fix save search timestamps so GTA V restores saved settings

Almo7aya:
Fix save searches returning a zero modification time, which causes some games to ignore existing saves and reset settings.
Report the newest file timestamp within each save directory, including subdirectories.

Verified in GTA V: graphics settings persist across multiple restarts.

### DIFF
--- a/src/libs/libSaveData.cpp
+++ b/src/libs/libSaveData.cpp
@@ -247,6 +247,18 @@ static std::string get_title_id() {
 	return title_id;
 }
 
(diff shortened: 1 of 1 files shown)


gandalfthewhite19890404:
Compiled PR does not allow to set language, that promt does not appears
image

Nmzik:
Credited. If saving fails(error in console), please remove the old save files.

PR #559, opened 2026-09-10
graphics: Fall back to hardware interpolation without VK_KHR_fragment_shader_barycentric

markl0101:
# Problem
GPUs without VK_KHR_fragment_shader_barycentric (e.g. GTX 10-series / Pascal) can't run the emulator. Confirmed issue on latest drivers. 

# Fix
Make barycentric support optional end to end, with hardware interpolation as the fallback. When barycentric shaders are unsupported, fallback interpolated value returned in its place. Notably, this may be an imperfect solution and will require further compatibility testing, but this at least allows for testing on previously unsupported GPUs.

# Testing
Verified in-game (Lego Brawls) on GTX 1070 Ti, previously threw error on launch. Theoretically works on other 10-series GPUs, untested.

# AI
Initial troubleshooting/research performed manually, with AI used to implement fix in all relevant locations.


rodjunger:
Duplicate of https://github.com/KytyPS5/KytyPS5/pull/551 ?

markl0101:
Yeah, have been testing for a few days and didn't realize another PR had gone up already. #551 appears to cover an additional fix as well, but in my testing did not resolve the barycentric shader issue on Pascal architecture (at least, in the game I've tested). Will test further this weekend and share feedback.

Hultwl:
@markl0101 
Looked through the diff — this is a more correct fix than what I did in #551. Mine only touches device selection / capability declaration and relies on Mesa ANV tolerating a declared capability that isn't formally enabled, so it doesn't survive NVIDIA's stricter validation. Yours actually changes SPIR-V codegen in spirvEmitterFlow.cpp (falls back to plain hardware interpolation, zeroes the bary-coord builtins), which is why it's the right direction for Pascal.

So, I put together a branch combining your barycentric fallback (cherry-picked as-is, full credit) with the depthBounds + input-stutter fixes from #551, since they're unrelated and don't conflict — figured one clean PR is better than two overlapping ones. Would you be up for that, and would you rather I open it as a new PR with you as co-author, or would you prefer to pull my depthBounds/stutter commits into your branch instead? Either way works for me, just want to avoid stepping on what you're doing.

markl0101:
@Hultwl thanks for digging into this! Single PR definitely cleaner, happy to be a co-author.

Hultwl:
Done. #586
i closed mine

Hultwl:
a small note, looks like main rewrote the files we were editing, so i also rewrote both fixes

it works on my UHD 620, can you test it with your GTX 1070 Ti?

markl0101:
Yeah sure, not at my PC right now but will test later today and confirm

PR #564, opened 2026-09-11
vscode: point IntelliSense at clang instead of MSVC

psnwd:
KYTY_SYSV_ABI expands to __attribute__((sysv_abi)) with no MSVC fallback, which is correct for a project that rejects cl.exe at configure time. Nothing told the C/C++ extension that, so it fell back to windows-msvc-x64 and flagged every guest-facing function in the tree with "__attribute__ is not a type name".

Set the IntelliSense mode, compiler path and C++ standard to match the compiler the build actually uses, and export a compile database for it to read.

## Summary by CodeRabbit

- **Chores**
  - Updated Visual Studio development environment settings for CMake.
  - Added default C/C++ IntelliSense configuration for Windows Clang x64, including C++20 support and compile commands integration.
  - Simplified environment path configuration.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,7 +10,11 @@
     "cmake.configureSettings": {
         "CMAKE_C_COMPILER": "clang-cl",
         "CMAKE_CXX_COMPILER": "clang-cl",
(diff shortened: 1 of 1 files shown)


Nmzik:
@psnwd all good except you hardcoded the path to VS

PR #578, opened 2026-09-11
graphics: pool the shader SRT evaluator memo storage

MehmetCambaz:
### Problem

Each SRT evaluator builds a private `std::unordered_map` memo for one evaluation and destroys it afterwards. In resource-heavy workloads this creates a large amount of allocator traffic.

In a fixed Kena: Bridge of Spirits scene, evaluator memo storage was reaching the system allocator about 87,800 times per frame.

### Change

Use a per-thread `std::pmr::unsynchronized_pool_resource` for evaluator memo storage, while keeping one private `std::pmr::unordered_map` per `Evaluator`.

- Memo contents are not shared between evaluators.
- Main, clean and `ReadFirstLane` child evaluators still have independent maps.
- Evaluation order and semantics are unchanged.
- Outputs such as snapshots, flattened SRT data and specialization state are not allocated from the pool.
- No evaluation-result caching is added.
- The pool retains its high-water mark until the evaluating thread exits. In the measured Kena workload this was about 0.3 MiB. If evaluation becomes multi-threaded in the future, retention would be per evaluating thread.

`Evaluator` is also made non-copyable so its polymorphic allocator cannot accidentally rebind to the default resource through a copy.

### Performance

Windows, fixed Kena scenes, alternating legacy/pooled blocks in the same process:

| Metric | Legacy | Pooled |
|---|---:|---:|
| `GetGraphicsPrograms` | 10.33 ms/frame | 8.89 ms/frame (-14.0%) |
| `GetGraphicsPrograms`, second scene | 10.43 ms/frame | 9.05 ms/frame (-13.3%) |
| Evaluator | 8.76 ms/frame | 7.32 ms/frame (-1.44 ms) |
| Mean frame time | — | roughly 2–3% better depending on scene |
| Evaluator memo allocations reaching the system allocator | ~87.8k/frame | 0 after warm-up |

All pooled blocks improved mean frame time relative to their neighbouring legacy blocks.

These numbers are from one title on Windows; equivalent gains are not claimed for other games or platforms.

### Validation

Added `srt_evaluator_memo_tests`, which runs the evaluator with pooled and unpooled memo storage and compares return status, outputs and reader-call order.

The tests cover, among other cases:

- separate main and clean evaluator contexts
- parent and `ReadFirstLane` child isolation
- nested `ReadFirstLane`
- evaluation re-entered from a reader callback
- two threads evaluating the same plan with independent readers
- thread exit and recreation
- allocation failure at every allocation point
- successful dependency memoization across enclosing failures
- preservation of 64-bit memo values and previously returned output

The focused suite passes 1,131 checks.

Full CTest result:

- current upstream: 36/37
- this change: 37/38

The only failure in both is the existing `kernel_file_system` Windows link failure.

A Kena gameplay smoke test with the production patch also completed without visual corruption or a crash.

### Portability

The implementation uses standard C++ PMR facilities:

- `std::pmr::unordered_map`
- `std::pmr::unsynchronized_pool_resource`
- `std::pmr::new_delete_resource`

It was built and tested with clang-cl and the MSVC standard library on Windows.

The project's current Linux/libstdc++ and macOS/libc++ toolchains support these facilities, but I did not build or benchmark this change on those platforms. Performance validation is Windows-only.

### Frame-pacing note

During the Kena A/B runs, pooled blocks also showed a small repeatable increase in frames landing near the ~100 ms vsync band: about +2.0 percentage points in one heavier scene and +0.4 points in a lighter scene.

This did not correlate with allocator activity or evaluator cost: warm pooled frames made no evaluator-memo system allocations, long frames did not show additional pool requests or an evaluator-time spike, while mean frame time and `GetGraphicsPrograms` consistently improved and P99 / >110 ms behaviour did not materially worsen.

I could not identify a causal mechanism in the pool for this observation.

AI assistance was used during investigation, implementation and testing. I reviewed the resulting change and validation before submitting it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -455,6 +455,17 @@ add_executable(resource_tracking_tests EXCLUDE_FROM_ALL
 target_link_libraries(resource_tracking_tests fmt::fmt)
 target_include_directories(resource_tracking_tests PRIVATE ${inc_headers})
 
+add_executable(srt_evaluator_memo_tests EXCLUDE_FROM_ALL
+	"${KYTY_TESTS_DIR}/SrtEvaluatorMemoTests.cpp"
+	"${KYTY_SOURCE_DIR}/graphics/guest_gpu/gpu_format.cpp"
+	"${KYTY_SOURCE_DIR}/graphics/host_gpu/hostMemory.cpp"
+	"${KYTY_SOURCE_DIR}/graphics/shader/recompiler/ir/passes/ResourceMaterialization.cpp"
+	"${KYTY_SOURCE_DIR}/graphics/shader/recompiler/ir/passes/SrtWalker.cpp"
+)
+target_link_libraries(srt_evaluator_memo_tests fmt::fmt)
+target_include_directories(srt_evaluator_memo_tests PRIVATE ${inc_headers})
+target_compile_definitions(srt_evaluator_memo_tests PRIVATE KYTY_SRT_TEST_HOOKS=1)
+
 add_executable(audio_out2_port_tests EXCLUDE_FROM_ALL
 	"${KYTY_TESTS_DIR}/AudioOut2PortTests.cpp"
 	"${KYTY_SOURCE_DIR}/libs/libAudio2.cpp"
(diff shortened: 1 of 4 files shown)


MehmetCambaz:
Superseded by #628, which includes this optimization and builds the dense memo work on top of it. Closing this one to keep the review in one place.

PR #582, opened 2026-09-12
shader: decode and translate V_FRACT_F16 (VOP1 0x5f) and V_CMPX_LT_U16 (VOPC 0xb9)

ADBHXNLOL:
Two gaps in the shader decoder tables, each of which aborted CFG construction for any shader using the instruction. Found while bringing up a commercial Frostbite title, which uses both heavily.

## VOP1 `0x5f` — `V_FRACT_F16`

`VOP1_OPCODE_LIST` jumps from `0x5e` (`V_RNDNE_F16`) straight to `0x60` (`V_SIN_F16`):

```
shader CFG build failed: unsupported decoded instruction in CFG at pc 0x000009b0:
unsupported family=VOP1 opcode=0x5f raw=[0x7e12bf07] reason=VOP1 opcode is not implemented
```

Decoding `0x7e12bf07` as VOP1 gives `vdst=v9`, `opcode=0x5f`, `src0=0x107` (v7) — `v_fract_f16 v9, v7`.

Everything needed already existed: `IR::ValueOpcode::FPFract32` lowers to the SPIR-V `GlslFract` extended instruction, and `Float16Unary` is the established path for this family. This adds the opcode to the enum, to both the VOP1 and VOP3-encoded VOP1 tables, to the float source/result predicates, to the SDWA rule table, and to the translator, mirroring `V_RNDNE_F16`. `invalid_negative` is `false` since negative inputs are valid for fract.

Without the SDWA rule entry, `FindVop1SdwaRule` returns `nullptr` and SDWA-encoded forms are still rejected, so that table needs the entry too.

## VOPC `0xb9` — `V_CMPX_LT_U16`

This one reports a misleading reason:

```
unsupported family=VOPC opcode=0xb9 raw=[0x7d72d4f9 0x86060000]
reason=VOPC SDWA modifier is not supported for opcode
```

The SDWA word decodes to `src0_sel=DWORD, src1_sel=DWORD`, a no-op selector, so the modifier is not the problem. `SupportsVopcSdwa()` is `opcode != Opcode::UNSUPPORTED`, so that message fires whenever the *opcode* failed to decode and the instruction happens to be SDWA-encoded. The real cause is that `0xb9` is absent from `VOPC_OPCODE_LIST`.

Identifying it needed care, because the table does not follow any single published AMD map. I reconstructed the numbering from the table's own entries — 8-wide blocks ordered `F, LT, EQ, LE, GT, NE, GE, T` — and checked it against every existing entry I could anchor: 23 agreements, 0 mismatches. The two `CLASS` ops are consistent, occupying the unusable always-false slot. The neighbouring `0xbc = V_CMPX_GT_U16` fixes the block base at `0xb8`, mirroring `V_CMP_*_U16` at `0xa8`, giving `0xb9 = V_CMPX_LT_U16`.

The translator reuses `EmitInteger16Compare` with `ULessThan32` exactly as `V_CMP_LT_U16` does, with the exec-writing flag set as for `V_CMPX_GT_U16`. One table entry covers both encodings, since `VOPC_OPS` backs the plain and VOP3-encoded lookups.

## Verification

- CI green on Windows, Linux and macOS.
- Confirmed in the built binary rather than only in the diff: `indices[0x5f]` in the VOP1 table goes from `0` — absent, so `LookupOpcode` returns `UNSUPPORTED` — to a valid entry index, and likewise `indices[0xb9]` in the VOPC table.
- In-game these two changes took the title from 1 compiled shader to 143 compute / 61 pixel / 45 vertex, past the recompiler entirely.

A formatting note: `VOPC_OPCODE_LIST` is laid out two entries per line, and I have put the new `0xb9` entry on its own line rather than repacking the surrounding rows, to keep the diff to a single added line. Happy to reflow it if you would prefer the table stay uniform.

While in the VOP1 table I noticed `0x59` (`V_FREXP_MANT_F16`) and `0x5a` (`V_FREXP_EXP_I16_F16`) are also missing. I have left them alone, since unlike fract they need real semantics written rather than a table entry, and I have no test case that exercises them.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ADBHXNLOL:
Heads up on two commits I pushed here that are not about the opcodes.

CI was failing on Windows with `-Wc++11-narrowing` errors in the SPIR-V backend — files this PR does not touch. `main` was red for the same reason, so it was pre-existing rather than something this branch introduced. I merged `main` in to pick up 3a27f8a, which fixed `SpirvBuilder.cpp`, but clang-cl still rejected the remaining sites, so the second commit covers those.

Constant operands like `spv::OpBitcast` narrow implicitly and are fine. The ones that fail are the non-constant cases: `EmitNative` forwards its argument pack straight into `AddFunction`, so `EmitGlsl` passing a `GLSLstd450` opcode as a runtime argument narrows, plus four `spv::Op` values chosen by a runtime ternary. I used `static_cast<uint32_t>` to match what you did in `SpirvBuilder.cpp`. `Unary`, `Binary`, `EmitCompareU32Constant` and `EmitBinaryU32` already take `uint32_t` opcodes, so I left them alone.

Happy to pull that build fix out into its own PR if you would rather keep this one to the decoder change — it only landed here because it was blocking the checks. Thanks for adding the test coverage.

ADBHXNLOL:
1208d22 supersedes the build commits I mentioned above — giving the builder typed operands that encode enums is clearly the right fix, rather than the per-site `static_cast` workarounds I had been adding.

Every conflict was in a file those commits touched, so I resolved each by taking your version and dropped my casts entirely. The branch is now back to just the decoder change plus your tests:

```
ShaderDecoder.h    |   2 +
VectorAluOps.cpp   |   8 +
Vector.cpp         |   4 +
tests/...          | 246 +
```

Sorry for the noise in the middle there — I was chasing the narrowing errors file by file while you were fixing the underlying cause.

PR #584, opened 2026-09-12
json2: export Value::clear

ADBHXNLOL:
Export sce::Json::Value::clear using the existing JsonValueClear payload cleanup. The wrapper preserves parent and root handler metadata, matching the SDK behavior for values that remain part of a tree.

The libSceJson2 NID FIjXN2TkuTs resolves to sce::Json::Value::clear().


Nmzik:
Value::setNullAccessCallBack - wrong implementation and kinda difficult to implement i think. Requires tests

PR #589, opened 2026-09-12
SpongeBob SquarePants (PPSA26893) boot fixes

YuhiAida:
## Summary

Takes **SpongeBob SquarePants: The Cosmic Shake (PPSA26893)** past its early boot crash (guest `abort()` about 1.7 s after graphics init).

### 1. `agc`: honor top bound for single-ended command buffers

The RHI thread calls `sceAgcDcbCopyData` on a temporary single-ended command buffer: `{bottom, top, cursor, 0, callback, 0, 0}` with `top - bottom` equal to exactly one CopyData packet. Kyty computed free space as `cursor_down - cursor_up`; with `cursor_down == nullptr` that is 0, so `ReserveDW` invoked the engine overflow callback, which produced `LowLevelFatalError [.../AgcRHI/Private/AgcSubmission.h] [Line: 442] Command buffer overflow.` and then the guest abort. `GetAvailableSizeDW()` now falls back to `top` when `cursor_down == nullptr` (restores the fallback from 1469c53).

### 2. `shader`: support `image_gather4_l` (MIMG opcode 0x44)

Adds the missing MIMG opcode. SPIR-V does not allow `Lod` on `OpImageGather` (the validator rejects it), so the LOD operand is currently dropped (level 0).

### 3. `shader`: model TTMP0-15 special scalar registers (108-123)

Extends the special scalar register count from 106 to 124 so TTMP0-15 are represented.

### 4. `launcher`: build with Qt below 6.6

`QVulkanInstance::NoPortabilityDrivers` only exists since Qt 6.6. The launcher now guards its use, so older Qt (verified with 6.4) builds too; no behavior change on the CI Qt (6.10).

## Testing

- Emulator builds and the game boots past the previous crash point (crash gone).
- No regression observed on an existing game run (DRAGON QUEST VII, ~90 s comparison).


YuhiAida:
image
there are still some issues after the first saving screen 
image

Nmzik:
@YuhiAida 
It’s fine. Your PR doesn’t need to be “perfect”. One thing to note: I remember when I introduced this fix to main in 1469c5304899126b8822ec24c4550922eb021eee, it regressed a bunch of games ( please retest if its still valid).

If you want, I can reopen the PR. Again, it doesn’t need to be perfect.

YuhiAida:
> @YuhiAida It’s fine. Your PR doesn’t need to be “perfect”. One thing to note: I remember when I introduced this fix to main in [1469c53](https://github.com/KytyPS5/KytyPS5/commit/1469c5304899126b8822ec24c4550922eb021eee), it regressed a bunch of games ( please retest if its still valid).
> 
> If you want, I can reopen the PR. Again, it doesn’t need to be perfect.

Allow me some more time, Since there are more issues I was thinking to make a bigger MR later

PR #590, opened 2026-09-12
shader: support IMAGE_GATHER4_L (MIMG 0x44)

francoisatt:
Problem. Rune Factory: Guardians of Azuma (PPSA21165) fails shader
recompilation on MIMG opcode 0x44 — see #281, raw=[0xf110010a 0x00c00d0a
0x00005e5f], still listed as current in the latest re-test.

Change. Decode 0x44 as IMAGE_GATHER4_L and route it through the existing
IMAGE_GATHER translation. The table entry carries ImageSampleFlagLod, so
ImageSampleAddressComponents accounts for the extra address register the
explicit-LOD form consumes. Nothing downstream changes.

Effect. The shader compiles. The explicit LOD is consumed but not applied:
the emitted instruction stays a plain OpImageGather, which samples at level
zero. Applying it would require the Lod image operand on OpImageGather,
which is only valid under SPV_AMD_texture_gather_bias_lod — a vendor
extension, not ratified, and not exposed by all drivers the project targets.
shadPS4 makes the same trade (its resource patching pass reads the LOD slot
and does not forward it to the gather opcode). A title that relies on
gathering from an explicit mip level will show an artifact rather than a
recompiler failure — I think that is the right trade here, but I'm happy to
revisit it if you'd rather reject the opcode than silently ignore the LOD.

Verification. Clean build on Windows, clang-cl 19.1.5, Ninja, based on
d3d7bd3. 38/39 tests pass. kernel_file_system does not link on clean main
either — undefined symbol: main, the same failure PR #565 addresses; see
also #505. Unrelated to this change.

The enum value is inserted inside the IMAGE_GET_RESINFO..IMAGE_GATHER4H
range that the uses_image check in the recompiler tests relies on. The
gather opcode table is built as a direct-index consteval table, so
inserting 0x44 ahead of 0x47 is order-independent.

Not verified at runtime — I don't have the title.

References
 - AMD RDNA 2 ISA Reference Guide, MIMG instructions: opcode 0x44
   IMAGE_GATHER4_L
 - Captured instruction word: 0xf110010a
 - SPV_AMD_texture_gather_bias_lod, on the Lod operand for OpImageGather

AI assistance. Written with AI assistance; every change reviewed, built and
tested locally by me.


francoisatt:
Heads-up on a duplicate: #613 also decodes MIMG 0x44, and its implementation is better than mine. My #590 only consumes the LOD register and leaves the gather at level zero — #613 builds the gather from the four samples at texel centres, so a shader walking a mip pyramid reads the right mip, and it fails loudly on the variants it can't verify rather than rendering wrong.

I'm happy to close #590 in favour of #613. The only argument for keeping mine is size — five lines in four files versus 32 files here — so if a reviewer wants the decode landed on its own first and the correct lowering separately, #590 is the smaller step. Otherwise it should go.

@psnwd your call on 0x44; I'd rather not have two PRs on it.

psnwd:
@francoisatt Go ahead with yours. Since it’s a much smaller change in terms of files and scope, I think it makes sense to land that first. My PR is still unfinished and has other scopes as well, so I’ll keep working on mine separately rather than closing it.

francoisatt:
Thanks — I'll keep #590 as is then.

One thing worth flagging for when you land the correct lowering: the gather path in spirvEmitterImage.cpp returns before the explicit-LOD handling, so layout.lod is allocated but never read for a gather. Your four-sample approach sidesteps that, but anyone extending the existing path will hit it.

I've also updated #281 to point Rune Factory at #613 rather than #590 — happy to flip that back now that this one lands first.

Nmzik:
credited

Issue #507, opened 2026-09-07
Resource tracking: support descriptor-table indices sourced from LoadBufferU32?


MehmetCambaz:
## Description

While testing GTA V PS5 (`PPSA04264`) on current upstream, Story Mode deterministically fails in shader resource tracking with:

```text
shader resource tracking:
hash=0x5e3e42fe237d3e16
stage=compute
pc=0x0000093c
GetBufferResource dword 0 is not a valid runtime value
```

The descriptor table itself resolves correctly. The failing part is the table index.

The relevant IR dependency chain is:

```text
ReadConstBuffer
  -> ShiftLeftLogical32
     -> ReadFirstLane
        -> SelectU32
           -> LoadBufferU32
```

The index is therefore derived from a buffer load in the vector domain and then made wave-uniform through `ReadFirstLane`.

Current `ValidateRuntimeValue` rejects the graph because `LoadBufferU32` is not one of the runtime-value forms it knows how to reconstruct.

## Reproduction

Tested on:

- GTA V PS5
- Title ID: `PPSA04264`
- content version: `01.010.002`
- upstream: `6aa6e390ee2930d7eeba7351c38bd342b783d933`

The failure reproduced 3/3 times with the same:

- shader hash
- compute stage
- PC
- descriptor dword

The game reaches the fully rendered main menu first, then selecting Story Mode eventually hits this failure.

## Relation to existing work

I also inspected the bounded-buffer/resource tracking work in #497.

Its `BoundedReadProof` appears to recognize the surrounding `ReadConstBuffer` / descriptor-table shape, but static inspection suggests the index eventually reaches `FiniteMaximum`, which has no `LoadBufferU32` case.

I could not test the relevant #497 commits cleanly on current upstream because #497 is based on an older resource-tracking implementation and current upstream has since refactored the validator.

So I do **not** want to assume that extending `FiniteMaximum` or allowing `LoadBufferU32` is necessarily the correct fix.

## Question

Is a descriptor-table index sourced from `LoadBufferU32` and then uniformized through `ReadFirstLane` intended to be supported by Kyty's current resource-tracking design?

If so, would the preferred direction be to extend the bounded-table/resource-materialization machinery for this case, or is there another architecture planned for runtime-loaded descriptor indices?

I can work on this if it fits the intended direction, but I wanted to confirm the design first rather than adding a speculative resource-tracking path.

fxpw:
[...]

PR #601, opened 2026-09-13
shader: implement S_MEMREALTIME (SMEM 0x25)

maximeallanic:
Ratchet & Clank: Rift Apart (PPSA01474, v01.005.003) stops during shader recompilation on a missing SMEM opcode:

```
ShaderRecompiler CS failed hash=0x46fa6901a17c4c27: unsupported decoded instruction in CFG
at pc 0x00000004: unsupported family=SMEM opcode=0x25 reason=SMEM opcode is not implemented
```

SMEM 0x25 is `S_MEMREALTIME`. Per the AMD ISA it "reads a 64-bit real time-counter and returns the value into a pair of SGPRS: SDST and SDST+1. The time value is from a constant 25MHz clock".

### What this does

* Decoder: `S_MEMREALTIME` in the SMEM table (2 dwords, no memory operands). The ISA only defines the result for an even SDST, so an odd SDST stays unsupported.
* IR: `GetRealTimeCounter(dword)` reads the counter, and the side-effecting `MarkRealTimeCounterRead` records that an invocation has read it.
* Bindings: two shader-data dwords are reserved after the packed memory offsets, only for shaders that use the counter.
* Renderer: `RebindBuffers` samples `KernelGetProcessTimeCounter` — the clock the title already observes — rescaled to 25 MHz with integer arithmetic, for every draw and dispatch.
* SPIR-V: the counter is read through `EmitShaderDataDwordLoad`, like user data.

### The part I would like your opinion on

The failing shader does not read the clock once. It reads the clock, reads a budget from its own buffer, and then loops re-reading the clock until the deadline passes:

```
start = realtime(); deadline = start + budget
while (realtime() < deadline) { work }
```

A Vulkan shader cannot observe time passing while it runs. With a counter sampled once per dispatch that loop never exits: the GPU hangs and the driver resets (`vkQueueSubmit failed: ErrorDeviceLost`, `nvlddmkm` event 153 on Windows).

So this models the counter inside an invocation: **the first read returns the sampled time and every later read returns `UINT64_MAX`**, which retires a deadline loop after one pass instead of spinning forever. It invents no duration, but a title does get one pass of work per dispatch where hardware would get many.

Other models exist (advance by an estimated per-iteration cost, or calibrate from the measured duration of the previous dispatch). I picked the one that cannot hang and invents no number, but this is a semantic decision and I would rather follow your preference than impose mine.

### Testing

Windows 11, RTX 4070, local build of `d3d7bd3`, PPSA01474:

| build | compute shaders compiled | result |
| --- | --- | --- |
| upstream | 11 | stops on the missing opcode |
| counter sampled per dispatch only | 17 | GPU hang, driver reset |
| this PR | 34 | no hang, no driver reset |

`shader_cfg_tests` passes, including a new case covering the decode, the IR lowering, the reserved shader-data dwords, the read marker and SPIR-V validation.

The title still renders nothing. With this change it stops later, in resource tracking, with `GetBufferResource dword 2 is not a valid runtime value` (`hash=0xf8bcd41f5f36626d`): that shader builds its whole buffer descriptor in VGPRs from a `BUFFER_LOAD_DWORD` and an emulated unsigned division, then makes it wave-uniform with `V_READFIRSTLANE_B32`. That looks like a second shape of #507, and I have not touched it here.

### AI disclosure

This change was implemented with AI assistance (Claude): reading the AMD ISA to identify the opcode, writing the patch, and analysing the logs and the failing shader. I reviewed the change, built it, and ran the shader tests myself on the machine described above, and I am responsible for what it does.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added support for shader real-time counter reads.
  * Real-time counter values are sampled consistently per shader invocation.
  * Added handling for the new real-time memory instruction across shader compilation and execution.
* **Bug Fixes**
  * Subsequent real-time counter reads now return a terminal value, helping deadline-based shader loops complete reliably.
* **Tests**
  * Added coverage validating decoding, intermediate representation, resource layout, and SPIR-V output.


Nmzik:
Credited

Issue #604, opened 2026-09-13
[GAME STATUS]: Zombie Cure Lab (Doesn't boot)


psnwd:
### Game title

Zombie Cure Lab

### Game ID / serial

PPSA22623

### KytyPS5 version

KytyPS5-2026-09-12-d3d7bd3

### Compatibility status

Doesn't boot

### Result details

Initialized: Config
Initialized: Log
Initialized: Timer
Initialized: Pthread
Initialized: Profiler
Initialized: Network
--- Build ---
Source build 1d28d3a
--- Error ---
could not reserve 13824 MB for guest direct memory. Windows commits
this up front, so the paging file is usually what needs to be larger.
KytyPS5\KytyPS5\src\kernel\memoryAddressSpace.inc:666

### Steps to reproduce the result

1. Open KytyPS5.
2. Boot the game.
3. If open normally then close it and try to re launch the game.

### Expected behavior

When you launch the game, then the intro cinematic begins. and show main menu.

### OS

Windows 11 Pro 25H2

### CPU

AMD Ryzen 7 7700X 8-Core Processor

### GPU

NVIDIA GeForce RTX 5070 Ti, driver 616.92

### RAM / VRAM

64 GB DDR5 RAM / 16 GB VRAM

### Log file upload

[_kyty.txt](https://github.com/user-attachments/files/32163117/_kyty.txt)

### Screenshots/video

<img width="860" height="482" alt="Image" src="https://github.com/user-attachments/assets/1435c822-f494-40cc-acab-88d17052aae4" />

Nmzik:
> could not reserve 13824 MB for guest direct memory. Windows commits
this up front, so the paging file is usually what needs to be larger.

Please retest. Increase page size, close other running application and restart

PR #605, opened 2026-09-13
kernel: implement _umtx_op timed waits

psnwd:
`UMTX_OP_WAIT` with a non-null timeout hit a hard `EXIT` stub, killing any game that
parked on a umtx with a deadline (Zombie Cure Lab, `PPSA22623`, does this during
startup).

Decode the relative `KernelTimespec` and convert it to microseconds, rounding
nanoseconds up so a sub-microsecond timeout does not collapse into an immediate
`ETIMEDOUT`. `SyncOnAddress::Wait64` takes a `uint32_t` microsecond timeout, which caps
at ~71 minutes, so longer waits are issued in slices and re-waited only while the result
is `ETIMEDOUT` and time remains. `tv_sec` beyond `INT64_MAX/1e6` saturates instead of
overflowing, and a `{0,0}` timeout returns `ETIMEDOUT` immediately to match FreeBSD.

The local result variable is named `wait_result` because `POSIX_CALL` expands to a
lambda that declares its own `result`.

The `_umtx_time` form (non-null `uaddr`, which carries `UMTX_ABSTIME` and an absolute
deadline) is still unimplemented and continues to hard-exit rather than silently
treating the deadline as relative.

Issue: https://github.com/KytyPS5/KytyPS5/issues/604

## Summary by CodeRabbit

* **Bug Fixes**
  * Timed waits now function correctly instead of being rejected as unsupported.
  * Timeout values are validated and handled accurately, including very long durations.
  * Wait operations now return the appropriate result when the timeout expires or the wait completes.


Nmzik:
Credited f3c82a317a002b41e0b89d451f3f608911c8be37

PR #613, opened 2026-09-14
Graphics shader fixes (Game: Beast of Reincarnation)

psnwd:
Fixes a set of shader and graphics faults hit by Beast of Reincarnation
(PPSA29343). Each one was a hard failure or silent no-op that blocked the
title from booting into gameplay.

### Shader recompiler

- Decode ttmp registers, `S_SWAPPC_B64`, `S_CODE_END`, `S_CBRANCH_CDBGSYS`.
  `S_CBRANCH_CDBGSYS` branches on the debug-mode bit, clear without a
  debugger, so it is decoded but kept out of the branch predicates: the CFG
  sees a plain instruction rather than an edge that can never be taken.
  `S_CODE_END` ends the executable body, so decoding stops there instead of
  running into metadata.
- `IMAGE_GATHER4_L` (MIMG 0x44) was undecoded and failed CFG construction.
  The shader using it walks a mip pyramid, so base-level sampling would read
  the wrong mip. `OpImageGather` takes no Lod operand, so the gather is built
  from the four samples it is defined to return, sampled at texel centres so
  the sampler's address modes still apply at edges. Unverifiable variants
  (array, cube, dref, offset) fail loudly instead of rendering wrong.
- 64-bit image atomics. RDNA2 picks atomic width via DMASK, not a separate
  opcode: two enabled components is one 64-bit value. Image is declared R64ui
  with a real 64-bit scalar sampled type (the IR's U64 is a u32 pair, which is
  a vector and not a legal sampled type), with bitcasts bridging the two. The
  descriptor check now requires the 64-bit guest format for 64-bit atomics
  instead of rejecting anything that is not k32UInt.
- Mesh groups run in either wave width. Entry code assumed wave64 for the wave
  index, lane base, per-wave caps and wave count, so wave32 was rejected. Also
  corrects the thread budget: the wave-count field is 4 bits, so the limit is
  15 waves at any width, not `15 * 64`.

### Graphics

- Stencil associations only redirect a stencil-shaped view. The redirect ran
  for any view at that address and returned before the depth-as-storage guard,
  so a colour storage binding on memory the guest had reused was handed a
  depth image. Later checks all test `info.pixel_format` while the view is
  built from the backing format, so nothing caught it.
- Block-compressed images get storage usage, so a compute pass writing raw
  blocks through an uncompressed view writes somewhere. They already carry
  `eExtendedUsage`; only the flag was missing. Layered images qualify only
  where maintenance6 reports `blockTexelViewCompatibleMultipleLayers`.
- Block-view descriptors size their extent in 4x4 blocks while MAX_MIP counts
  the texel chain. Block level N still covers texel level N, so the range is
  clamped to the levels that exist rather than renumbered. Renumbering selects
  a mip too large. Multisampled descriptors pass through untouched.
- Enable `VK_EXT_shader_image_atomic_int64` and `VK_KHR_maintenance6` when
  present, as optional extensions, so devices without them are still selected.
- Present through an sRGB swapchain, and keep the guest surface's format in
  the prepared frame so the blit converts once.

### Audio

- Raw AAC channel configuration comes from the stream. `uiMaxChannels` is what
  the port can emit, not what the stream carries, so declaring it makes a
  stereo stream fail on its first channel pair element. Zero is not an option
  either: it needs a program config element the two-byte header has no room
  for. Candidate layouts are tried until one decodes, then locked.

### AGC

- Patch end-of-pipe action packets built after the address, and return an
  error code instead of aborting on an unsupported layout.

### Testing

Built with clang-cl. All test suites pass: shader_cfg, shader_recompiler_compute,
resource_materialization, resource_tracking, image_page_table,
shader_vertex_metadata, scalar_provenance. Verified against the title with
Vulkan and shader validation enabled: no validation errors, no SPIR-V failures.

### Known remaining

3D renders correctly into the HDR scene buffer, but the tonemap output contains
different contents, so that pass samples the wrong image. Not addressed here.

## Summary by CodeRabbit

* **New Features**
  * Added support for 64-bit image atomic operations when supported by the graphics device.
  * Added support for additional shader instructions, trap-temporary registers, explicit-LOD texture gathers, and variable wave sizes.
  * Improved handling of compressed textures with multiple layers.
  * Added support for sRGB swapchain presentation without unnecessary format conversion.

* **Bug Fixes**
  * Improved stencil-image association handling and invalid image-view diagnostics.
  * Improved AAC channel-layout detection and resilience to corrupt packets.
  * Improved shader handoff and termination handling.
  * Queue end-of-pipe patching now accepts zeroed packet templates.


psnwd:
Update:
Seems like it's getting good results. It just needs a bit more work, especially with Unreal support (Nanite, UMG blur, etc.). We also need to work on the audio.

image

PR #616, opened 2026-09-14
fix: PRX symbol resolution and user ID aliasing

bipinkrish:
### What Changed:

* **Threads & Kernel Synchronization**:
  * Removed global condition variable waiter tracking (`g_cond_waiters`) and lock contention in `threads.cpp`.
  * Replaced `CondVar::SignalThread` with a no-op to eliminate glibc broadcast/cancellation deadlocks during signal delivery.

* **Dynamic Linker & Symbol Resolution**:
  * Added support for 1-part (`name`) and 2-part (`name#lib`) ELF symbols in both `CreateSymbolDatabase` and `RuntimeLinker::Resolve` for PRX and middleware loading.
  * Optimized `sceKernelDlsym` (`kernel_find_export_symbol`) by validating Base64 NID charsets, fast-tracking direct NID lookups, and eliminating duplicate linear scans.

* **Controller & User Service**:
  * Unified user ID validation via `Config::IsPrimaryUserId`, allowing standard dummy IDs (`0`, `1`, `100`) across both `PadOpen` and `libUserService`.
  * Added monotonic timestamp fallback (`KernelGetProcessTime()`) in `pad_fill_data` to prevent initial 0-timestamp glitches.

---
  
Nearly every bug fixed directly corresponds to how the Unity engine is architected on PlayStation 5, Identified these when i was testing Slit game (Silt 01.003.000 PPSA03306) after the fixes it went from Menu Screen to In Game running at 30FPS in Linux

[output_10mb.webm](https://github.com/user-attachments/assets/5ac14aeb-433e-4797-8f1b-074e8d291e32)

---

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring the audio input device and AMD CPU features.
  * Improved support for primary user profiles, including reserved system user IDs.
  * Enhanced symbol resolution for library-qualified names, versions, and valid identifiers.

* **Bug Fixes**
  * Improved controller and user-service behavior for supported user profiles.
  * Corrected handling of missing controller timestamps.
  * Improved fiber switching, signal handling, and condition-variable wake-ups.
  * Added broader support for AMD instruction emulation across platforms.


Nmzik:
I'm pretty sure we don't need to mount /devlog (PS5 debug stuff left in code?) Correct me if I am wrong :). Thanks for optimizations, I will take a look

bipinkrish:
@Nmzik you are right, i removed mounting  /devlog

bipinkrish:
@Nmzik thanks for taking the condvar fix (and the co-author credit). What would you like done with the rest of this PR?

Nmzik:
@bipinkrish 
Nice optimization ideas. Looks like glue for now.
Slit should go in-game and shouldn’t get stuck at random times. I don’t want to close PR for now

bipinkrish:
@Nmzik sounds good,  I’ll rebase onto main, keep the remaining bits.

PR #634, opened 2026-09-16
input: add gyroscope and motion sensor support

dylanriverajw138-stack:
Capture gyroscope and accelerometer samples through SDL controller events, retaining each sample for buffered pad reads. Convert acceleration to G, preserve gyro values in radians per second and their axis directions, and integrate timestamped gyro samples into a normalized relative orientation.

Wire motion enable/disable and orientation reset to controller state. Preserve motion during UI input resets and skip integration across long sensor-report gaps. The controller data layout stays unchanged; advanced motion filters remain unimplemented.

Validation: Linux emulator build and CLI smoke test passed. A focused test harness passed checks for units, axes, sample history, quaternion rotations and composition, reset, enable/disable, reconnection, report gaps, reserved bytes, and concurrent reads. Three independent audits completed, including a final compactness and cleanliness review. Physical controller and gameplay testing is pending.


Nico00007566:
I hope it will be added soon. I want to run some tests with gyro.

PR #640, opened 2026-09-16
shader: implement software BVH ray intersection for MIMG 0xe6

carbonimax:
Shaders using `IMAGE_BVH_INTERSECT_RAY` (`MIMG 0xe6`) currently stop at the unsupported-instruction path. This adds software RTIP 1.1 BVH intersection through the existing GPU buffer-device-address path, without requiring Vulkan ray-tracing extensions.

The supported variant uses 32-bit node pointers and triangle return mode 1. It handles triangle nodes 0/1, FP16/FP32 boxes, barycentric remapping, box growth and child sorting, with bounds checks and EXEC masking. Unsupported packed/64-bit variants and descriptor modes remain explicit failures; `0xe7` is decoded but not lowered.

The captured shaders also require two related compiler corrections:

- Keep loop-dependent scalar buffer reads and raw GPU data pointers on the GPU instead of evaluating them as CPU-side resource-table snapshots. Preserve descriptor validation and buffer bounds checks.
- Do not classify branches to loop exit tails as direct structured loop exits. Use the existing dispatcher fallback when the graph cannot be structurized. A previous CFG-only test accepted invalid SPIR-V; its replacement verifies preserved topology, instruction coverage, Phi parents and SPIR-V validity.

Validation:

- Native builds succeeded, including the GPU test executable; 9/9 targeted tests passed, covering shader CFG, scalar provenance, resource materialization/tracking and the existing macOS/CLI regressions.
- All 27 shaders captured during a local launch of the game pass `spirv-val --target-env vulkan1.3`, including all four BVH shaders. The game reaches the separate missing-mesh-shader support check.
- CPU IR execution tests cover triangle and box results, sorting, bounds, null descriptors, inactive lanes and address carry. A synthetic GPU triangle test is available through `shader_recompiler_compute_tests --bvh-only`.
- The isolated PR branch applies cleanly to current `main`; changed C++ translation units pass syntax checks and the project clang-format hook passes.

GPU BVH execution is **not yet verified**: the four captured BVH dispatches are zero-sized, and the synthetic GPU test stops before dispatch because the existing Vulkan test harness requires image-gather/compute-derivative features unavailable through the tested MoltenVK setup. This PR does not claim correct rendered output or playable game support.

The build, native-test and game-launch results above were obtained in the integrated macOS checkout with #633, #635 and #637 applied. Those fixes are not included in this standalone PR; a full build and runtime run of the rebased standalone branch have not been repeated.


francoisatt:
Ran this on Windows with a native Vulkan driver (RTX 5080, clang-cl,Ninja) — the part of the harness your MoltenVK setup can't reach.

Good news first: --bvh-only passes. MimgBvhIntersectRayTriangle runs tocompletion on the GPU, so the synthetic triangle test does execute andsucceed on real hardware.

The full suite is a different story. shader_recompiler_compute exits0xc0000409 on:  ScalarMemoryLoadVariants failed at dispatch: BDA test has no explicit    guest mapping

On clean origin/main (63fb822) the same test passes — I rebuilt and re-ran to check. Caveat: your branch is based on 18a1e0f, and main has since touched SrtWalker.cpp, ResourceTracking.cpp and added ~2100 lines to ShaderRecompilerComputeTests.cpp, so I can't say whether it passed on  *your* base. What I can say is that it passes on main today and crashes on this branch.

Given the change keeps loop-dependent scalar buffer reads on the GPU as BDA reads, the message reads like the test's guest mapping isn't set up for that path. Might be the test needing an explicit mapping rather than a bug in the lowering.

Everything else: 35/39 ctest, the other three failures are the known SDL link problem (kernel_file_system,  irtual_memory_allocation, guest_red_zone_patcher — see #565, present on clean main too).

Happy to re-run once it's rebased on current main.

carbonimax:
Thanks for testing this on native Vulkan and confirming that `--bvh-only` passes on the RTX 5080.

Updated in a588023: merged current `main` (63fb822) into this branch and fixed the missing guest-memory mappings in three compute tests:

- `ScalarMemoryLoadVariants`
- `ScalarLoadSignedImmediateOffsetAddsSoffset`
- `ScalarLoadAlignsComponentsAndMasksAddress`

This PR keeps data-only scalar address loads on the GPU instead of collecting them as CPU SRT reads. These fixtures therefore need an explicit BDA mapping. Each now maps guest address 0 to the start of its initial buffer; the shader instructions and expected results are unchanged. A host-side audit of all `MakeCases()` shaders found these three missing mappings; the other BDA cases already declared theirs.

Also fixed the undefined 32-bit shifts in the BVH test evaluator reported by CodeRabbit, masking the shift count with `& 31` to match the IR semantics.

Local validation: x86_64 Release build passed; `shader_cfg`, `scalar_provenance`, `resource_tracking`, and `resource_materialization` passed. The new portable `--scalar-memory-host-only` selector passes for all three fixtures, including SPIR-V validation and mapping checks. GPU execution remains unverified here: the MoltenVK harness stops before dispatch because image gather / compute derivative quad support is unavailable.

For a native Windows retest, the new selector isolates the three affected cases:

```powershell
.\shader_recompiler_compute_tests.exe --scalar-memory-only
.\shader_recompiler_compute_tests.exe --bvh-only
ctest --output-on-failure
```

Could you rerun these on the updated branch? The host checks do not establish that the full GPU suite now passes.

francoisatt:
Re-ran on 3cb78d5 (Windows, native Vulkan, RTX 5080, full clean build).

Two of the three fixtures are fixed: ScalarMemoryLoadVariants and ScalarLoadSignedImmediateOffsetAddsSoffset both pass on GPU now.
--bvh-only still passes.

The third one now fails differently — no longer a crash, a wrong result:

    ScalarLoadAlignsComponentsAndMasksAddress failed at readback:
    expected [0x11111111, 0x22222222] actual [0x22222211, 0x22222222]

The first dword comes back as one byte of word 0 followed by three bytes of word 1, so the read looks byte-offset by 3 rather than snapped to the dword boundary. Given the fixture's name, the address mask doesn't seem to be applied on the BDA path.

Worth knowing: --scalar-memory-host-only reports all three ok, including this one. So the host-side selector doesn't catch it — it's a host-vs-GPU divergence, which is exactly the gap your MoltenVK harness can't reach.

Rest of the suite unchanged: 35/39, the other three failures are the known SDL link issue (#565), present on clean main too.

carbonimax:
Thanks for the GPU retest. Fixed in 5233c5a9b2de76acac6263632c0ae949a38b336f: the scalar BDA path aligned the offsets but left the guest base unaligned and its upper address bits unmasked. It now aligns the base to four bytes and masks it to 48 bits before adding the aligned offsets. Added a SPIR-V dataflow regression check for ScalarLoadAlignsComponentsAndMasksAddress; removing either the base alignment or the upper-bit mask makes it fail. Build, shader_cfg_tests, and --scalar-memory-host-only pass locally. I could not validate the native GPU result here. Could you rerun ScalarLoadAlignsComponentsAndMasksAddress (expected [0x11111111, 0x22222222]), the other two scalar-load fixtures, and --bvh-only on Windows/native Vulkan with your RTX 5080?

francoisatt:
Re-ran on 5233c5a (Windows, native Vulkan, RTX 5080, full clean build).

All three scalar-load fixtures now pass on GPU, including ScalarLoadAlignsComponentsAndMasksAddress (the byte-offset readback is gone).
--scalar-memory-host-only: 4/4 ok
--bvh-only: MimgBvhIntersectRayTriangle ok

Full suite: 36/39. The three remaining failures are the known SDL link issue (#565): kernel_file_system, virtual_memory_allocation, guest_red_zone_patcher, all "Not Run", same as on clean main.

From my side the GPU path is clean now.

theantipopau:
Tested #640 with the real game on Windows + AMD, since the description says GPU execution wasn't verified and the GPU runs reported so far were on NVIDIA.

**Setup:** tested PR head 5233c5a merged onto upstream `main` f100f78 (clean merge, builds). I checked the current head (b584ee5, at the time of writing) and it's the same 5233c5a plus just a rebase onto newer `main` — the BVH-related source files (RayTracing.cpp, DynamicBuffer.cpp, ImageOps.cpp, ShaderCFG.cpp) are byte-identical, so this result should still hold. Windows 11, Ryzen 7 9800X3D, Radeon RX 9070 XT, AMD driver 26.8.1 (LLPC), Vulkan 1.4.349. Astro Bot PPSA21564, no patches.

**Before (plain main):** exits after ~12 s with `unsupported family=MIMG opcode=0xe6` (raw=[0xf1989f07 0x00040505 0x4442413d 0x4543403e 0x00004746]), same as #281.

**With #640:** the decode error is gone and the BVH compute shaders now translate, but the game never gets past frame ~125-163 (title bar: fps 0) and one thread stays pinned at 100% CPU for 15+ minutes. Sampling that thread with the build's PDB shows it is inside the AMD driver, not the emulator:

```
CommandProcessor::DispatchIndirect
 -> RenderExecutor::DispatchDirect
  -> PipelineCache::GetComputePipeline
   -> CreatePipelineInternal   (Libs::Graphics)
    -> [AMD driver, no symbols]  (vkCreateComputePipelines)
```

With `--graphics-debug-dump true`, the dump timestamps show the driver taking longer on each BVH kernel:

| kernel (SPIR-V size) | driver compile time |
|---|---|
| cs_503d7f066a496c3c (610 KB) | ~3 min |
| cs_1f16e50eea0c89e3 (649 KB) | ~6 min |
| cs_ff8ee744ffa4dcdc (683 KB) | >6 min, still compiling when I stopped |

`spirv-val --target-env vulkan1.3` passes on the dumped kernels, so this doesn't look like invalid SPIR-V. Structurally each has a single `OpLoopMerge` + `OpSwitch` (the dispatcher fallback from the CFG change), ~1050 function-scope `OpVariable`s, ~710 `OpPhi`s and ~230 `OpFunctionCall`s (~35k disassembled lines). I have not tested whether the structured (non-dispatcher) path avoids this, or whether NVIDIA behaves the same. I'm not attaching the dumps because they're derived from a commercial game's shaders; the hashes above are for cross-reference.

Happy to try a variant (e.g. avoiding the dispatcher fallback for these kernels) if you want another AMD data point.

*AI disclosure: I used Claude Code to help build, run the game, sample the stack and analyze the SPIR-V. I ran and checked everything on my own machine.*

PR #655, opened 2026-09-17
Decode image_gather4_l, TTMP0-15 and the s_swappc_b64 front-program hand-off

YuhiAida:
# Decode image_gather4_l, TTMP0-15 and the s_swappc_b64 front-program hand-off

Three small, independent decoder/recompiler gaps that each stopped SpongeBob
`PPSA26893` shader compilation completely. Each commit fixes one measured boot
stop and moves the title to the next one; each carries a targeted test case.

Split out of a much larger local branch so each change can be reviewed on its own.
No new configuration switches, no behaviour changes outside the shader frontend.

## Commits

| # | Commit | Fix | Measured before -> after |
|---|---|---|---|
| 5 | `shader: support image_gather4_l (MIMG opcode 0x44)` | decode MIMG `0x44` (`image_gather4_l`) instead of failing the CFG build | `shader CFG build failed ... family=MIMG opcode=0x44 at pc 0x484` -> shaders compile: VS 12 / PS 14 / CS 83 (was 0) |
| 6 | `shader: model TTMP0-15 special scalar registers (108-123)` | accept scalar codes 108-123 as SGPR-class operands (`ttmpN`), `NumScalarRegs` 106 -> 124 | `unsupported scalar source operand 0x00000073 at pc 0x1c0` -> compiles (next stop is the front-program walk) |
| 4 | `shader: handle the S_SWAPPC_B64 front-program hand-off` | treat `s_swappc_b64` as a front-program terminator, not only `s_setpc_b64` | `unknown RDNA2 instruction family at pc 0x23c, raw=0xd2f2c70f` -> MS phase shader decodes 335 instructions, CFG 34 blocks, SPIR-V 29469 words; VS 16 / PS 20 / CS 122 / GS 1 |

Diff: 9 files, +97 / -6. `tests/shaderCfgTests.cpp` carries the new coverage
(`TestNewShaderRecompilerImageGatherVariants` extension, TTMP scalar-operand cases,
`TestFusedShaderSwapPcHandoff`).

## Why one PR

All three are the same class of defect in the same component (frontend decode /
front-program walk), they do not interact outside that component, and the title's
boot path needs all three to get past shader compilation at all - so they cannot be
tested independently end to end. Kept as three separate commits so each can be
reviewed (and reverted) on its own.

## Known limitations

- `image_gather4_l` carries an explicit LOD, but core SPIR-V `OpImageGather` has no
  Lod operand, so the LOD is dropped and level 0 is sampled (the same path the
  existing LZ variant uses). The decoded dump still names the opcode
  `image_gather4_l`, and the emitted module validates. Nothing in this title's
  current boot path distinguishes it; this is an ISA-level approximation that would
  need a real mip-level gather consumer to observe.
- `NumScalarRegs` growth (106 -> 124) only affects SSA arrays and bounds checks;
  no baseline counts change.

## Verification

- Branch is `origin/main` (`63fb822`) + these three commits, cherry-picked from the
  local branches where each fix was validated. The nine touched files are
  byte-identical (git blob hash) to the verified branch state.
- `shader_cfg_tests`: rc=0 (includes the three new cases).
- Repro: SpongeBob `PPSA26893` boot, same run script and timeout per step, comparing
  the boot-stop signature and the `Shaders:` compiled counts per stage.

## Not in this PR

- The renderer depth `expand-clear` fix (`depthRenderTarget.cpp`) that unblocks the
  next stop is a separate renderer change and will be its own PR.
- The VRAM/device-budget item that surfaces after that is also separate.

## Summary by CodeRabbit

* **New Features**
  * Added support for decoding and handling `image_gather4_l` shader instructions.
  * Added support for `s_swappc_b64` instructions and temporary scalar registers.
  * Improved shader program termination and register display for mesh-shader handoffs.

* **Bug Fixes**
  * Improved compatibility with shader programs using image gather operations and fused mesh-shader transitions.

* **Tests**
  * Added regression coverage for temporary registers, image gather decoding, and mesh-shader handoffs.


Nmzik:
@YuhiAida I think it’s in-game. Please check on your side. You were credited.

Strangely enough, I don’t remember any shader changes related to TTMP0–15 special scalar registers.

YuhiAida:
> @YuhiAida I think it’s in-game. Please check on your side. You were credited.
> 
> Strangely enough, I don’t remember any shader changes related to TTMP0–15 special scalar registers.

Thanks
Not in-game yet; it stops in the renderer, not the shader frontend:

validation on: fatal VUID-VkWriteDescriptorSet-descriptorType-00339 (an image bound as a storage image without VK_IMAGE_USAGE_STORAGE_BIT).
validation off: VS 125 / PS 159 / CS 203 / GS 2 compiled, ~635 frames, then image.cpp exits creating a 3840x2160 image - device heap 0 at 4.43/4.48 GiB (98.7%).

So my blocker right now is device memory (resident set / VMA budget), not shaders.

TTMP0-15: it came from a measured stop on a pre-104530e base - a shader read ttmp7 and the frontend refused it (unsupported scalar source operand 0x73 at pc 0x1c0). On 104530e I cannot reproduce it (489 shaders compiled, zero such decode stops, memory exit first), and the other two commits here are already covered by your tree. So I'll close this PR.

I'll look at the device-memory stop next and, if I get it working, send it as a separate PR.

PR #670, opened 2026-09-17
emulator: validate numeric CLI arguments

c0sxm0s:
## Summary
--screen-width, --screen-height, --gpu, and --vblank-frequency were parsed with
Common::ToInt32, which ignores std::from_chars errors and returns 0, so values
like "--screen-width abc" or "--gpu xyz" were silently accepted: zero screen
size, or GPU 0 selected instead of auto (-1). Reject invalid values with an
error message like the other options, using from_chars-based helpers
(ParseUint32/ParseInt32); reject zero screen dimensions.

## Test
- `--screen-width abc|0`, `--screen-height -5`, `--gpu xyz`,
  `--vblank-frequency foo` now fail with "invalid ..." and exit code 1
- `--screen-width 1920` and `--gpu -1` still parse and proceed
- Windows, clang-cl 23.1.1, Qt 6.12.0: kyty_emulator builds clean


Nmzik:
Lets keep it simple. User is responsible for passing valid arguments

PR #672, opened 2026-09-17
common: fix swapped min/max in KYTY_ENUM_RANGE; drop unused EnumName8

c0sxm0s:
## Summary
`KYTY_ENUM_RANGE(e, mx, mn)` assigned its max parameter to
`magic_enum::customize::enum_range::min` and its min parameter to `max`,
inverting the configured range for any caller. Assign them in the order the
parameters are documented. The macro currently has no callers, so this fixes
it before it gets used.

Also removes `EnumName8`, an unused exact duplicate of `EnumName`.

## Build
Windows, clang-cl 23.1.1, Qt 6.12.0: kyty_emulator builds clean.


Nmzik:
Credited

PR #675, opened 2026-09-17
feat: Add basic Discord Rich Presence integration

1joepie1:
### Description
Adds lightweight Discord Rich Presence integration via Windows Named Pipes (`\\.\pipe\discord-ipc-0`) without adding external heavy dependencies.

### Changes
- Added `src/launcher/discord_rpc.h`
- Updated `src/main.cpp` to initialize presence on startup.


1joepie1:
Going to fix this ASAP am new to the discord API

1joepie1:
If you have any ideas in how to improve L me now.

Almo7aya:
Not sure if we should make it inside the launcher, I think it should be part of the cli emulator, and it should be controlled on/off, by a param

1joepie1:
> Not sure if we should make it inside the launcher, I think it should be part of the cli emulator, and it should be controlled on/off, by a param

I hope its fixed now

1joepie1:
> Not sure if we should make it inside the launcher, I think it should be part of the cli emulator, and it should be controlled on/off, by a param

Oh and do you want the parameter as an setting in app or more like an config file

PR #678, opened 2026-09-17
kernel: fix inverted bit pattern in KernelClearEventFlag

c0sxm0s:
sceKernelClearEventFlag clears the specified bits from the event flag, but Clear kept only the given bits and cleared all others (m_bits &= bits instead of m_bits &= ~bits), corrupting the flag for any game that calls it. The internal ClearMode::Bits path in Wait already used the correct semantics.


Nmzik:
This is NOT a bug

PR #659, opened 2026-09-17
Persist SaveDataMemory2 save data to disk (fixes Astro Bot progress not saving)

defektu:
## Summary
- Astro Bot (PPSA21564) uses the `SaveDataSetupSaveDataMemory2` / `SaveDataGetSaveDataMemory2` / `SaveDataSetSaveDataMemory2` / `SaveDataSyncSaveDataMemory` API instead of the classic `SaveDataMount3` / `Prepare` / `Commit` flow. That whole family only ever touched an in-process `std::vector`, so game progress (coins, level state, rescued bots) was silently discarded on process exit — every boot looked like a new game, even though the game itself was calling these functions correctly.
- `SaveDataSetupSaveDataMemory2` now loads an existing `_SaveData/<title>/sce_sdmemory_<slot>.bin` from disk into the memory buffer on first setup, and `SaveDataSetSaveDataMemory2` / `SaveDataSyncSaveDataMemory` now flush the buffer back to that file instead of being a no-op.
- Enables `PRINT_NAME` logging for the module and adds `LOGF` tracing (call params, mount path existence, per-write offsets, and which branch/result each mount call took), since none of this was previously visible and made diagnosing the missing persistence much harder than necessary.

## Test plan
- [x] Built `kyty_emulator` target (Release, clang-cl + Ninja) with the change.
- [x] Ran Astro Bot (PPSA21564): rescued a bot, jumped ahead a level, confirmed `_SaveData/PPSA21564/sce_sdmemory_0.bin` is written with real XML save state (`NumCoin`, `PlayingLevelId`, level `IsCleared`/`IsUnlocked`, `Bot/<id>/IsRescued`).
- [x] Relaunched the title and confirmed progress carried over across the restart (coin count and rescued-bot flag persisted from the previous run instead of resetting).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #696, opened 2026-09-18
savedata: persist save data memory durably

1OO1O11O:
## Problem

`sceSaveDataSetupSaveDataMemory2` / `Get` / `Set` / `Sync` are backed by one process-wide `std::vector<uint8_t>` that is never written to disk:

- progress is lost the moment the emulator exits — the save data memory API is how a number of titles store their progress;
- every user, slot and title shares the same buffer, so they overwrite each other;
- `Get` returns a zeroed `SaveDataParam` and an icon of size 0 regardless of what was stored;
- `SaveDataSyncSaveDataMemory` does nothing and never queues the completion event a title waits on.

## What this does

Adds `Libs::SaveData::SaveDataMemoryStore` (`src/libs/saveDataMemory.{h,cpp}`) and wires the four entry points to it. Each title/user/slot keeps its own snapshot at:

```
_SaveData/<title id>/sce_sdmemory[N]/<user id>/memory.sdm
```

**Durability.** Guest bytes, `SaveDataParam` and the icon live in one checksummed container, so an update is published by a single rename: write a temporary file → flush it to the platter → rename over the previous save → flush the directory entry. Interrupt it anywhere before the rename and the previous save is still intact; a file that does not decode is reported as `SAVE_DATA_ERROR_BROKEN` instead of being handed to the title as its save. I/O failures are mapped to save data errors and returned, never swallowed into a successful-looking save.

**Write-through.** Every `Set` is published before it returns, so a title that never calls `sceSaveDataSyncSaveDataMemory` still keeps its progress across an exit or a crash. `Sync` re-publishes (a retry for a previously failed durability barrier) and, in its asynchronous form, queues `SAVE_DATA_EVENT_TYPE_MEMORY_SYNC_END`.

**Windows sharing conflicts.** A replacing rename fails while any other process holds a handle that does not permit deletion — which is exactly what a virus scanner or search indexer does for a moment after every file close. Uncontended, this lost roughly two out of three saves on the machine this was developed on. Such a failure is now waited out for about a second before the save is reported as failed; a genuine permission error still surfaces.

**Isolation.** A slot is held under an exclusive lock file while it is set up, so a second emulator process cannot write the same save concurrently (it gets `SAVE_DATA_ERROR_BUSY`). The event queue gained its own mutex, since the sync event is now queued from a path that already holds the mount lock.

**API fidelity.** `init_param` / `init_icon` are only read when the caller passes `SET_PARAM`, since a title that did not ask for them may leave those pointers uninitialized; setup never truncates existing progress or overwrites existing metadata with initialization data; `user_id < 0` returns `SAVE_DATA_ERROR_INVALID_LOGIN_USER`; `data_num` is capped at the documented maximum; the store stamps `param.mtime` on every update; a `Set` with nothing to update does no I/O. The title id is reduced to characters that cannot form a path component of their own before it names a directory (CWE-22) — deliberately not a `CUSA`/`PPSA` format check, which would leave homebrew and system titles unable to save at all.

## Tests

`tests/SaveDataMemoryTests.cpp` (registered as `save_data_memory`, added to the Windows, macOS and Linux CI jobs) covers:

- parameter validation, unsafe title ids, and ordinary non-`CUSA` title ids;
- `SET_PARAM` gating of the initialization data;
- isolation across users, slots and titles, including cross-process exclusion;
- injected I/O failures at create, write, flush, rename and directory-flush, each checked for "the old save is still there and still complete";
- corrupt, truncated and extended save files;
- a scanner-style sharing conflict (fails without the retry above);
- concurrent writers with no lost updates;
- child processes killed mid-write, mid-flush, mid-rename and after publication, with a fresh process verifying that what it reads back is either the complete old snapshot or the complete new one.

`tests/save_data_memory/CMakeLists.txt` builds the same test standalone, so it runs on a machine with no Vulkan SDK, Qt or glslangValidator:

```
cmake -S tests/save_data_memory -B _Build/save-memory -G Ninja
cmake --build _Build/save-memory && ctest --test-dir _Build/save-memory --output-on-failure
```

Ran green 25+ consecutive times on Windows (MinGW). The full emulator build was not exercised locally — this environment has no clang-cl/Vulkan toolchain — so the CI jobs are the real check on that; `src/libs/libSaveData.cpp` was compiled standalone against the emulator headers.

## Relationship to #659

#659 fixes the same bug with a different on-disk layout (raw `memory.dat` plus a `param.bin` sidecar) and is the older PR; this one focuses on making an update atomic and its failures visible: a single checksummed file replaced by one rename, flushes checked, and crash/interruption behaviour covered by tests that kill real child processes. The container here is deliberately named `memory.sdm` so neither implementation can read the other's file as if it were its own format. Happy to rebase on top of #659, fold the two together, or close this if you prefer that approach.


Nmzik:
closed in favor #659. If there is something we missed, feel free to open new PR/new bug report

PR #712, opened 2026-09-19
tests: link KernelFileSystemTests on Windows

diegoalejandrogomez:
## Summary

`kernel_file_system_tests` does not link on Windows:

```
lld-link: error: undefined symbol: main
>>> referenced by D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:78
>>>               msvcrtd.lib(exe_main.obj):(int __cdecl invoke_main(void))
```

The test includes `SDL.h`, which pulls in `SDL_main.h`. On Windows that header defines `SDL_MAIN_AVAILABLE` and renames `main` to `SDL_main`, expecting the `SDL2main` library to provide the real entry point. The test target only links the static SDL2 library, so no `main` exists. The emulator itself never hits this because `src/main.cpp` does not include `SDL.h`.

This defines `SDL_MAIN_HANDLED` before the include on Windows, which is the opt-out documented in `SDL_main.h` for programs that own their `main`. `SDL_SetMainReady` is not needed: with `SDL_MAIN_AVAILABLE` (as opposed to `SDL_MAIN_NEEDED`) the ready flag in `SDL.c` defaults to true.

## Test plan

- [x] Windows, clang-cl + Ninja (Debug): `kernel_file_system_tests` builds and links.
- [x] Running the binary reaches the socket cases (it previously could not be built at all).

## Notes

- `kernel_file_system` is not in the ctest filter for any platform in `build.yml`, which is why this went unnoticed. Adding it to the Windows job would keep this from regressing; left out of this PR as a CI policy call.
- On Windows the binary currently aborts at `guest PEEK and WAITALL preserve the wake bytes`. That is a separate pre-existing Winsock issue (`recv` rejects `MSG_PEEK | MSG_WAITALL` with `WSAEOPNOTSUPP`); fix to follow in its own PR.


diegoalejandrogomez:
This was already solved by this #565 
Closing this pr since is duplicated

PR #715, opened 2026-09-19
graphics: prototype native GPU PerVertex replay and shader caches

carbonimax:
The opt-in PerVertex path now lowers vertex/fragment shaders in native C++, runs vertex unpack/capture on the GPU, and replays the captured per-corner values. It also persists transformed SPIR-V and Vulkan driver pipeline caches.

A concrete failure fixed here was Metal compilation reporting undeclared `buffers` and `buffersBufferSize`: reassembling source SPIR-V 1.3 as 1.5 requires resource and private globals in `OpEntryPoint`. Capture and fragment interfaces are rebuilt accordingly, and transformed modules are validated before caching.

### Scope
- Dynamic capture/replay layouts, bounded GPU vertex/index unpacking, and descriptor/push-constant integration.
- Checksummed bytecode cache with semantic layout checks and build-time transformation source signatures; development driver-cache reuse and atomic saves.
- GPU regression covering storage-buffer descriptor arrays, `OpArrayLength`, push constants, capture/replay output, and cache corruption.
- Disabled by default; enable with `KYTY_PER_VERTEX_PROTOTYPE=1`. No Python subprocess or CPU vertex readback in this path.

### Validation on the isolated branch
Based on upstream `main` at `8d66c4e`. The live checkout and its remaining WIP were left separate.

```sh
cmake --build _Build/macos --target kyty_emulator shader_cfg_tests per_vertex_transform_tests per_vertex_gpu_execution_tests -j 8
_Build/macos/shader_cfg_tests
_Build/macos/per_vertex_transform_tests
SDL_VULKAN_LIBRARY=/path/to/libMoltenVK.dylib _Build/macos/per_vertex_gpu_execution_tests
```

- All four targets build on macOS (x86_64/Rosetta).
- `shader_cfg_tests`: exit 0.
- `per_vertex_transform_tests`: exit 0.
- `per_vertex_gpu_execution_tests`: exit 0 on Apple M4 Pro / MoltenVK 1.4.3, including SNORM16 signed endpoints and negative intermediate normalization, strict synthetic GPU comparisons, and production cache-loader corruption cases.
- Optional captured game shaders are not distributed with the repository and were explicitly skipped in this clean checkout.
- Repository formatting hooks and diff checks pass.

### Review scope and remaining limitations

This PR is ready for review as an experimental, disabled-by-default native PerVertex implementation and cache foundation. It is not a claim that the game renders correctly or is playable.

The supported draw contract remains deliberately restricted: triangle lists, one vertex buffer, at most eight attributes, Float32/Float16/Unorm decoding, SNORM16x4 decoding, and additional explicit guards. Unsupported cases fail explicitly instead of dropping draws or fabricating values.

Windows, macOS and Linux CI passed on the previous head `7d17838`. The Windows compatibility fix accounts for the legacy MSVC STL's implicit multiline regex behavior; a CPU parser regression runs on all three CI platforms. Current base is upstream `main` at `8d66c4e`; Windows, macOS, and Linux CI passed for `7fe93a8` ([run 35508197398](https://github.com/KytyPS5/KytyPS5/actions/runs/35508197398)).

A bounded integration run passed the previously rejected SNORM16x4 attribute guard and continued through additional PerVertex draws. It then stopped on an independent vertex-shader translation limitation. The live checkout contains other WIP, so this run is evidence for passing the format guard, not validation of this isolated branch or visual correctness.

SNORM16x4 is now implemented in the CPU guard and generated GPU unpack shader, with CPU and MoltenVK fixtures for `INT16_MIN`, `INT16_MAX`, zero, and a negative intermediate value. The bounded run still does not establish correct rendering or playability.

Follow-up work is to diagnose the independent vertex-shader translation limitation before resuming the persistent black-screen investigation. Persistent-cache growth and startup cost observed in the live checkout also remain separate; the existing microbenchmarks do not establish an end-to-end launch-time improvement.

Related work remains separate: #633 (optional host features), #640 (BVH), #643 (centroid), #685 (MoltenVK descriptor policy), #721 (device-address buffer lifetime), and #727 (NGG/mesh). This PR does not include those changes, the mip-view experiment, or video changes.


carbonimax:
Follow-up `7d17838` adds the exact PerVertex input format observed in the game: `0x47` (`k16_16_16_16Float`), four 16-bit components at attribute offset 20 with stride 28.

The change keeps the scope to Float16 only, generates the unpack SPIR-V from checked-in GLSL, and covers signed zero, subnormal, normal, infinity, and NaN conversion on CPU plus finite Float16 fixtures on MoltenVK.

Validation on an isolated checkout of this PR:
- Release x86_64 build: `kyty_emulator`, `shader_cfg_tests`, `scalar_provenance_tests`, `per_vertex_gpu_execution_tests`
- CPU suites: passed
- MoltenVK GPU suite: passed
- `spirv-val --target-env vulkan1.2`: passed

A bounded integration run reached GPU completion of the first PerVertex draw without DeviceLost. That runtime proof also included other open integration changes, so it does not by itself qualify this PR in isolation. The black screen remains and the later subgroup-function failure is a separate issue.

PR #717, opened 2026-09-19
graphics: clamp host mip levels and gate layered block views

YuhiAida:
## Summary

Running Dragon Quest VII Reimagined (PPSA17942) with `VK_LAYER_KHRONOS_validation` enabled takes
the emulator down twice during early gameplay, with a fatal validation error each time:

- `VUID-VkImageCreateInfo-mipLevels-00958` — the guest can declare a mip tail deeper than the
  host image's mip chain allows, so `vkCreateImage` is rejected.
- `VUID-VkImageViewCreateInfo-image-09487` — a block-format (BC) view with `layerCount > 1` is
  only legal with the maintenance6 `blockTexelViewCompatibleMultipleLayers` capability.

Both abort through `EXIT`, so they are fatal, not cosmetic. This PR removes both and keeps the
information needed to see where guest mip tails get mapped.

## What changed

- `image.{h,cpp}`: `ImageOps::HostMipChainLength` (`bit_width(max(w,h,depth))`) and
  `ImageOps::HostMipLevels` (`min(guest levels, chain)`); the host image is created with the
  clamped count. Barriers, resolves, copies and image-to-image transfers map a phantom guest
  range onto the deepest host level instead of referencing a level that does not exist.
- `ImageOps::ReportMipClamp` logs one line per site (capped at 32 prints) when a range is
  clamped: address, type, format, extent, guest/host level counts, range, layers, samples,
  pitch, tile mode. Source and destination ranges are reported independently, so an out-of-chain
  source cannot hide behind a valid destination range.
- `textureCache.cpp`: uploads drop regions above the host chain, `ClearImage` clamps to the
  deepest host level and derives `full_image` from the backing count, and
  `SynchronizeBufferFromImage` declines volume mip tails rather than reading levels the host
  image does not have.
- `imageView.cpp`: view mip ranges are clamped the same way. A multi-layer uncompressed view of
  a block-compressed image is rejected before creation when the device lacks the capability,
  since the layer count cannot be legalised without maintenance6 and the driver would reject or
  misread the view.
- `vulkanWindow.cpp` / `graphicContext.h`: `VK_KHR_maintenance6` is enabled and
  `VkPhysicalDeviceMaintenance6Properties::blockTexelViewCompatibleMultipleLayers` is queried
  at device creation, with the result logged.

Nothing changes for images whose guest mip count already fits the host chain — `HostMipLevels`
returns the guest count in that case.

## Evidence

With Vulkan validation, SPIR-V validation and GPU-AV enabled:

| | before | after |
|---|---|---|
| result | aborts with `VUID-VkImageCreateInfo-mipLevels-00958` at guest `00:00:25.158` | 900 s run, guest `00:12:09`, still presenting |
| validation errors | 1 fatal (plus the follow-up) | 0 |
| clamp diagnostics | – | 2 create / 15 view / 15 barrier reports, all inside guest `00:00:24.69–25.24` |

The clamp reports land in exactly the window where the unfixed build died, and the post-fix run
carries none of the previous `[Vulkan][E]` output. The capability line reports
`Vulkan block texel view multiple layers support: true` on the development GPU, so the layered
block view is created natively there; the rejection path covers devices that report `false`.

A later 120 s run of the current head with validation enabled and no `vk_layer_settings.txt`
present (no VUID muting) reports 0 errors and only the pre-existing `out_mrt_0` draw warning
(10 lines, the layer's duplicate-message limit), so the result does not depend on muting any
VUID.

Gameplay reaches and passes the second stop on hardware — that is the only confirmation for the
`09487` site, since the pre-fix abort makes an automated A/B of that exact content point
impractical.

## Tests

- `kyty_tests` builds.
- Host-side (GPU present): `image_page_table`, `gpu_tiler`, `texture_cache_layered_image`,
  `host_image_allocation`, `texture_cache_image_views`, `texture_cache_storage_sampled`,
  `texture_cache_depth_readback`, `compute_meta_clear_classification`, `buffer_cache_dirty_gc`
  all pass. (The graphics tests need a real device — they abort with `ErrorIncompatibleDriver`
  in the build container.)
- CLI tests (`emulator_user_name_cli`, `emulator_user_name_too_long`,
  `emulator_present_mode_cli`, `emulator_present_mode_invalid`) pass in the container.

## Notes for reviewers

- Phantom guest levels are *mapped*, not emulated: no data is stored for a mip level the host
  chain does not have, and transfers that would need it are skipped or declined (volume mip
  tails), with a report line. If a title is later found to genuinely depend on reading such a
  level, that case needs a real fix in the guest mip layout handling rather than this clamp.
- `ImageOps::ReportMipClamp` is a diagnostic; it is intentionally not gated behind a config
  flag, but is capped at 32 lines per site so it cannot flood the log.


YuhiAida:
**Outside-diff finding "Do not mark the destination GPU-modified when no copy occurs" (`textureCache.cpp:697-700`) — addressed in cae4e77.**

Took the suggested shape: `Image::CopyMip` now returns whether it actually copied (false when `mip >= destination.backing.mip_levels`, i.e. the guest mip maps beyond the host mip chain), and `TextureCache::CopyImageMip` marks the destination GPU-modified only on success, so `ResolveDepthOverlap`/`ExpandImage` can no longer claim ownership of pixels that were never transferred.

While tracing the callers I found the same silent return in `Image::Resolve`, and it was worse in that case: `ResolveColorTargets` called `cache.MarkGpuWritten(dst.image_id)` *before* the resolve, and `CommitGpuWrite` clears `BufferModified`/`CpuDirty`, so a skipped resolve would additionally mask a missing blit. `Resolve` also returns bool now and the mark only happens after a successful resolve. The remaining clamp sites (`view`, `barriers`, `clear`) clamp onto the deepest host level and proceed, so they were left as-is.

On reachability: `ReportMipClamp` had a single global 32-report budget which the `view`/`barriers` reports were saturating, so the other sites' real counts were invisible. It now has a per-site budget (first 4 verbatim, then a log-scale heartbeat with a total). A 120 s replay of the DQ7 sequence with `VULKAN_VALIDATION=true` on the fixed build reports `copy-mip=0`, `resolve=0`, `view≥32`, `barriers≥32`, `create=3` — so on this content both skip paths are latent rather than currently corrupting state, and the run finished with no fatal or validation errors. The guard is there for when a guest image does present a mip beyond the host chain on another path.

PR #724, opened 2026-09-20
Zarchive Loading

Aspenini:
Adds the ability to load Zarchive files, made from folder dumps of PS5 games. This allows games to be stored as a single file with a bit of compression. It is the same format that shadPS4, Cemu, and Xenia use for their consoles' games.

Tested on Dead Cells, which loaded the same as a folder and .zar file.

AI was used to assist development.


Heliumrich:
Amazing, I imagine this will work well with updates/DLCs as other .zar archive files, loaded ontop ?
And what about modding game files, can we load a loose folder ontop ?

fab918:
Could you provide the file size and loading time figures before and after compression for a few games?

Crispy81:
Be interesting to see how much space is saved and also if there's any overhead ie - longer processing/boot times/CPU usage, etc...

Heliumrich:
That's just classic Zstandard (zstd), if there are uncompressed files it can save quite a lot, but usually what takes the most storage space are textures, and they are barely compressible, so it will probably save between 1-10% for 95% of games.

There is a tiny CPU overhead yes, but with modern (multicore) CPUs it's probably not even detectable through benchmark noise.

For me the best thing about this is the rare times I need to copy dumps over to an external SSD or whatever ; +200k files through your SSD controller (and worse, USB controller) slow things down a lot...

PR #732, opened 2026-09-20
auto-hide cursor on idle and start (window)

### LINKED ISSUE #282: Cursor auto-hide on game window
Auto hide the cursor on the game window when the game starts, it can be shown on mouse movement trigger and auto hide on mouse idle, it helps to avoid confusion with games that uses in-game cursors like Blue Prince.

anatoliikuc:
Hides the mouse cursor over the game window when the game starts, and brings it back the moment you move the mouse. If it sits still for a couple seconds it hides itself again. Should help with games that draw their own cursor (e.g. Blue Prince), where the OS cursor just gets in the way.

Fixes #282


src/graphics/presentation/window/window.cpp
```diff
@@ -806,10 +859,12 @@ void WindowContext::Run() {
...
 
-		if (!HostInputWaitEvent(&loop.event)) {
+		if (!HostInputWaitEvent(&loop.event, CursorAutoHideWaitMs())) {
+			UpdateCursorAutoHide(CursorTimeS());
 			continue;
 		}
 		ProcessEvent(timer.GetTimeS());
+		UpdateCursorAutoHide(CursorTimeS());
 	}
 }
```
davidrohusch, line 869: The whole mouse visibility function sounds more like an event to process.

Making the main loop more complex when it can be avoided is something I would preferably not do.

I subjectively think it would be more fit to extend `ProcessEvent()` 

Any mouse motion could just invoke a `SDL_AddTimer(CURSOR_IDLE_HIDE_SECONDS * 1000, CursorHide() ) ` + CursorShow()

davidrohusch: Oh okay, nevermind. AddTimer is not desirable as it would spawn another thread, SDL prefers the graphic calls to be done within main thread. This would not work.

PR #751, opened 2026-09-21
net: resolve guest PEEK and WAITALL flags on Windows

### LINKED ISSUE #505: kernel_file_system test fails on main: guest PEEK and WAITALL preserve the wake bytes

rudy-07:
Closes #505

Under FreeBSD / POSIX, combining MSG_PEEK and MSG_WAITALL is valid. On Windows Winsock, passing both MSG_PEEK and MSG_WAITALL to recv() fails with error WSAEOPNOTSUPP (10045).

In ConvertMessageFlags(), when peeking on Windows, clear MSG_WAITALL so that host recv()/recvfrom() peeks at the socket buffer without error, preserving the wake bytes as asserted in KernelFileSystemTests.

Also define SDL_MAIN_HANDLED ahead of SDL.h in KernelFileSystemTests.cpp so the test executable links on Windows without SDL_main rewriting.


rudy-07:
Thanks for the feedback! I have updated the PR (commit 1361216):

1. MSG_WAITALL Emulation on Windows: Rather than returning short reads on partial arrivals, Recvfrom now emulates MSG_WAITALL when combined with MSG_PEEK on stream sockets by awaiting host_len pending bytes via FIONREAD / select / WSAPoll. It also properly handles remote peer disconnection (POLLHUP / EOF) by returning partial accumulated data per POSIX semantics, and respects SO_RCVTIMEO.

2. Docstrings: Added Doxygen documentation for ConvertMessageFlags and Recvfrom to satisfy docstring coverage checks.

francoisatt:
Tested on Windows (native build, clang-cl, full clean build), as the reporter of #505. Since the CI workflow is still awaiting approval, this may be the only Windows run so far.

Same base for all three builds (ba55ba5, the merge-base of this PR):

| Build | kernel_file_system |
|---|---|
| main | does not link (`undefined symbol: main`, SDL) |
| main + only the SDL_MAIN_HANDLED change from this PR | fails: `0xc0000409`, "guest PEEK and WAITALL preserve the wake bytes" (the #505 failure) |
| this PR (1ce0fc8) | passes |

So both parts are needed: the SDL define makes the test link, and the network.cpp change fixes the actual PEEK+WAITALL failure.

Full suite on this PR: 39/40 as run, the 40th (save_data_memory) was only missing from my local build script; built separately, it passes.

Not verified: the regression test sends the whole payload before receiving, so it does not exercise the WAITALL emulation added in 1361216 (partial arrival, peer close, SO_RCVTIMEO), nor the nonblocking skip from 1ce0fc8. That is the case Copilot raised. Also, the PR description still describes the first approach (clearing MSG_WAITALL), which may be worth updating for reviewers.

rudy-07:
Thank you so much for taking the time to test all three configurations natively with clang-cl and confirming the fix! That validation is super helpful.

Regarding your observations:

1. Regression Test Coverage: I've pushed commit 007c68c, adding explicit test assertions in KernelFileSystemTests.cpp for both:
     - Non-blocking sockets: Verifies that empty receive with 0x42 returns EWOULDBLOCK immediately when ORBIS_SO_NBIO is set.
     - Chunked partial arrival: Verifies that a stream socket receiving payload in chunks ("wa" followed by "ke\0") waits for the full length before returning under 0x42.

2. PR Description: Updated the PR description to accurately detail the MSG_WAITALL emulation, non-blocking bypass, error handling, and test coverage for reviewers.

francoisatt:
Heads-up: #797 (theantipopau) covers the same Winsock MSG_PEEK | MSG_WAITALL limitation and the same SDL_MAIN_HANDLED fix, with a broader implementation (accepted sockets, datagrams, MSG_DONTWAIT, EOF) that also handles the partial-arrival case Copilot raised above. Both are open and neither references the other. My test results above still stand for this PR; posting this so it gets resolved rather than both stalling.

